PR #14, opened 2019-02-20
Use mypy to check the code

thomkeh:
* add mypy.ini
* add run_mypy.py
* add check_all.py
* run run_mypy.py on Azure

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,3 +33,6 @@ jobs:
 
   - script: python run_pylint.py
     displayName: 'pylint'
+  - script: |
+      python -m mypy ethicml
+    displayName: 'mypy'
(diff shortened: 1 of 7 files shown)


olliethomas:
Does mypy cause the build to fail if it finds something?

thomkeh:
> Does mypy cause the build to fail if it finds something?

Yes, for the moment. We can wrap it in a script like you did with pylint to not fail on every error.

olliethomas:
That might be nice, though at first glance mypy doesn't return a score, which made the pylint script trivial.

thomkeh:
Tests pass now

olliethomas:
I guess we’ve already done the test to demonstrate the build fails if mypy flags anything

thomkeh:
> I guess we’ve already done the test to demonstrate the build fails if mypy flags anything

Yes, the first time it failed because mypy found an error.

thomkeh:
I didn't remove `python -m pytest tests/` from `azure_pipelines.yml` in this PR.

Issue #9, opened 2019-02-15
Proposal: algorithm in separate processes


thomkeh:
I imagine running algorithms in separate processes to work something like this:

At the end of the data preparation, the data is stored on the hard drive:

```python
with TemporaryDirectory() as tmpdir:
    tmp_path = Path(tmpdir)
    data_path = tmp_path / Path("data.parquet")
    df = pd.concat([data_dict['x'], data_dict['s'], data_dict['y']], axis='columns')
    df.to_parquet(data_path, compression=None)
```

Then a python script for the algorithm is started as a separate process. The path to the data file is passed as a commandline argument.

```python
from subprocess import call
PYTHON_EXE = "/home/ubuntu/anaconda3/envs/pytorch_p36/bin/python -u"

...

cmd = f"{PYTHON_EXE} {script_path} {data_path}"
call(cmd, shell=True)
```

Then the data file is read by the called script.

```python
import sys
import pandas as pd

...

data_path = sys.argv[1]
with open(data_path, 'rb') as f:
df = pd.read_parquet(data_path)
```

(`parquet` has to be install with `pip install fastparquet`.)

thomkeh:
More info on Parquet:

> Parquet is designed to faithfully serialize and de-serialize `DataFrame` s, supporting all of the pandas dtypes, including extension dtypes such as datetime with tz.

> Several caveats.

> * Duplicate column names and non-string columns names are not supported.
> * Index level names, if specified, must be strings.
> * Categorical dtypes can be serialized to parquet, but will de-serialize as `object` dtype.
> * Non supported types include `Period` and actual Python object types. These will raise a helpful error message on an attempt at serialization.

http://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#io-parquet

olliethomas:
I think this looks like a good idea. If we want to be able to have many jobs over the HPC then this will become necessary so best to get it in early

thomkeh:
this has been started in #17 . closing this.

PR #17, opened 2019-02-20
Run algorithms in their own process

thomkeh:
See #9

### DIFF
--- a/ethicml/algorithms/algorithm_base.py
+++ b/ethicml/algorithms/algorithm_base.py
@@ -12,3 +12,12 @@ class Algorithm(ABC):
     def name(self) -> str:
         """Name of the algorithm"""
         raise NotImplementedError()
+
+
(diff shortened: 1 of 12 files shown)


thomkeh:
The tests had been failing because I had hard-coded the path to the python executable for the standalone logistic regression:

```python
python_exe="/Users/tk324/anaconda/envs/pytorch/bin/python"
```

which obviously doesn't work on Azure. So now I'm using `sys.executable` instead which is basically perfect for this task:

> **sys.executable**
> A string giving the absolute path of the executable binary for the Python interpreter, on systems where this makes sense. If Python is unable to retrieve the real path to its executable, `sys.executable` will be an empty string or `None`.

https://docs.python.org/3/library/sys.html

---

The API of threaded algorithms is not perfect yet, but I think this can be merged already.

Issue #16, opened 2019-02-20
Proposal: replace the dataframe dictionary with a NamedTuple


thomkeh:
Dictionaries always have the problem that you can store anything in there and you can also store nothing in there. `dictionary['x']` can always fail.

NamedTuple on the other hand have a fixed number of fields and all fields are mandatory:

```python
from typing import NamedTuple
import pandas as pd

class DataTuple(NamedTuple):
    x: pd.DataFrame
    y: pd.DataFrame
    s: pd.DataFrame
```

They are instantiated as follows:

```python
data = DataTuple(x=df1, y=df2, s=df3)
```

The values are accessed like this:
```python
data.x
```

Finally, NamedTuple are *immutable*. This: `data.x = df4` throws an error.

olliethomas:
Then the type of data will be `data: NamedTuple[str, pd.DataFrame]` ?

thomkeh:
Done in #22

PR #22, opened 2019-02-21
Replace the dataframe dictionaries with a datatuple

thomkeh:
See #16


thomkeh:
The definition of the new type is in `algorithms/utils.py`:

```python
class DataTuple(NamedTuple):
    """A tuple of dataframes for the features, the sensitive attribute and the class labels"""
    x: pd.DataFrame  # features
    s: pd.DataFrame  # senstitive attributes
    y: pd.DataFrame # class labels
```

PR #23, opened 2019-02-25
Improve handling of threaded algorithms

thomkeh:
* improve error handling in case the called script fails
* abstract a lot of ThreadedLR into its parent classes
* move common functions out of `examples/logistic_regression.py`
* add a threaded SVM model


thomkeh:
This only touches code that was added in #17 so I'll merge this.

PR #24, opened 2019-02-25
Make it easier to import the metrics

thomkeh:


thomkeh:
There is not really anything objectionable here I think. I'll merge now.

PR #30, opened 2019-02-28
Remove make_data_tuple()

thomkeh:
I should have removed it when I introduced `DataTuple`s.


olliethomas:
This is a good move - much clearer

PR #37, opened 2019-03-05
WIP - thought for refactoring threaded algos to fit the existing API

olliethomas:
I think the work on threading might be able to be re-factored so that the user doesn't need to know the inner workings. 

What I mean is instead of 
```
model = ThreadedModel()
result = run(model, train, test)
```

we can do 
```
model = Model()
model.run(train, test, sub_process=True)
```

as a halfway house i've done this so that the default for sub_process is False, but I think this could (and should) change to default=True.

This has a couple of nice benefits, the user should feel no difference to the API, we can tidy the code base a little by pulling methods into the Abstract Classes and we don't need to have any duplication.

Putting this as a draft because if agreed it's the start of a longer process, there's still some stuff to work out. 1) How to pass init params for models like Beutel 2) Pulling out Threaded versions of abstract classes as there's already quite a few places they're used. Both are do-able, but maybe we could pair to go through them?


olliethomas:
Interesting that it fails on azure, but seems fine on my local. I'm going to park this until you're back @thomkeh so we can chat about it. It looks like this might tie in nicely with (or completely break) what you're doing pulling the models into their own modules

thomkeh:
The error is "No module named 'ethicml'". I'm assuming this is because we don't install ethicml on Azure. We could do `pip install .` in the CI script and it would probably work, but I'm not sure that is the correct way to do it. Maybe we shouldn't use `__module__`, but I'm not sure.

thomkeh:
I like this refactoring. My only worry is imports.

I still have this ideal in my mind where you don't need to install PyTorch in order to use EthicML. This means that the Beutel model should not be automatically imported when you import ethicml.

The good news is that this should still be possible to achieve with your new API. Basically, we need to remove the import of `beutel` in `ethicml/algorithms/preprocess/__init__.py`. This means that the Beutel model will only be imported on request. So, the following will **not** work anymore:

```
import ethicml
model = ethicml.algorithms.preprocess.beutel.Beutel()
```

But what **will work** is:

```
from ethicml.algorithms.preprocess import beutel
model = beutel.Beutel()
```

i.e. `beutel` has to be specifically requested.

I *think* this will work as well, but I haven't tested it:

```
from ethicml.algorithms.preprocess.beutel import Beutel
model = Beutel()
```

 Scipy is doing the same thing I think. You can't do `import scipy as sp` and then do `sp.optimize.curve_fit()`. You have to do `from scipy import optimize`.

ethicml/algorithms/inprocess/svm.py
```diff
@@ -5,17 +5,32 @@
...
+from ethicml.algorithms.inprocess.in_algorithm import InAlgorithm
+from ethicml.algorithms.utils import DataTuple
 
 
 class SVM(InAlgorithm):
     """Support Vector Machine"""
-    def run(self, train: DataTuple, test: DataTuple) -> pd.DataFrame:
+    def run(self, train: DataTuple, test: DataTuple, sub_process: bool = False) -> pd.DataFrame:
+        if sub_process:
+            return self.run_threaded(train, test)
```
thomkeh: It's a bit ugly that this has to be repeated everywhere...

We could have a wrapper function like this:

```
def run(..., sub_process):
    if sub_process:
        self.run_threaded(...)
    else:
        self.__run(...)
```
and the real implementation would be in `__run()`.

ethicml/algorithms/inprocess/in_algorithm.py
```diff
@@ -35,3 +41,34 @@ def run_test(self, train: DataTuple, test: DataTuple) -> (
...
+        """ runs algorithm in its own thread """
+        pred_path = tmp_path / "predictions.parquet"
+        args = self._script_interface(train_paths, test_paths, pred_path)
+        self._call_script(str(ROOT_PATH / "algorithms" / "inprocess" / self.filename), args)
+        return self._load_output(pred_path)
+
+    @staticmethod
+    def _load_output(file_path: Path) -> pd.DataFrame:
+        """Load a dataframe from a parquet file"""
+        with file_path.open('rb') as file_obj:
```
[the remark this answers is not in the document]
olliethomas: no, the problem is the return type. In the pre-algortithm class the _load_output function is different

ethicml/algorithms/inprocess/svm.py
olliethomas: yeah, might need to have a think about what's the most friendly way of doing this. So if we decided to add FairGP, then what would be a reasonable set of methods to ask to be implemented

olliethomas:
So, I've got it so that Beutel is running with parameters... But I think it needs to be made more generic. It is reasonable that we start passing hyper-parameters to (for example) the SVM, but at the moment we only have flags in the pre-process base class. 
I'm going to leave this as WIP for now unless we want to do work on the other draft PR and it would help if master was updated

PR #38, opened 2019-03-08
attempt 1 at switch to pipenv

olliethomas:
Seeing if my understanding of pipenv matches up with Azure's understanding


olliethomas:
A bit of background on pipenv https://realpython.com/pipenv-guide/
and the official docs https://pipenv.readthedocs.io/en/latest/

olliethomas:
It looks to me as though pipenv will potentially be a solution to running models in their own venvs. So I thought it would be a good idea to get familiar with it by moving ethicml away from a requirements.txt and to a Pipfile

olliethomas:
Closing as this was part of the sep_envs commit

PR #42, opened 2019-03-08
Make it optional to use self.__module__

thomkeh:


thomkeh:
Not needed anymore with #52 .

PR #43, opened 2019-03-12
Separate environments for algorithms

olliethomas:


olliethomas:
SHHHHHHHOOOOOOUUUUULLLLLDDDDDDD be good to go ;)

olliethomas:
Notes for @thomkeh 
- Azure now uses python2 not pypy. I think this was because python2 wasn't installed on the base image, so something went wrong and it was defaulting to pypy.
- The problem with feather is fixed, I hadn;t deleted the failed clone of the repo on my local test environment making it look like there was a problem when there wasn't.
- The problem now is that pandas and pygame aren't being installed properly. I'm not sure why so latest run has pygame removed from the SVMTWO repo to see what happens.

olliethomas:
ok, so latest is removing pygame worked.... but that's a bit of a cheap hack. the whole point of it being there is to ensure that the internal venv is running, not the outer one for the entire project. I'll try putiing it back and seeing if it breask again. If so, we can reason that it's because the internal venv isn't being used as expected

olliethomas:
now removed the pipfile.lock on the imported repository

PR #46, opened 2019-03-12
Add isolated algorithms based on pipenv

thomkeh:
Same as #43 but this PR doesn't depend on #38


thomkeh:
It's still using pypy...

thomkeh:
#43 was merged; so this is not needed anymore.

PR #62, opened 2019-03-27
Add Fair GP model

thomkeh:


thomkeh:
It's missing tests, but I'm not sure we want to add a test for this because it will slow the tests down again.

PR #63, opened 2019-03-28
Example

olliethomas:


ethicml/evaluators/evaluate_models.py
```diff
@@ -79,49 +80,68 @@ def evaluate_models(datasets: List[Dataset], preprocess_models: List[PreAlgorith
...
-        train, test = train_test_split(load_data(dataset))
-        if test_mode:
-            train = get_subset(train)  # take smaller subset of training data to speed up training
+    columns = ['dataset', 'transform', 'model', 'repeat']
+    columns += [metric.name for metric in metrics]
+    results = pd.DataFrame(columns=columns)
 
-        to_operate_on = {"no_transform": {'train': train,
-                                          'test': test}}
+    total_experiments = 1 + repeats*(len(datasets) * (len(preprocess_models)+1)) + \
```
thomkeh:
```suggestion
    total_experiments = len(datasets) * repeats * (1 + len(preprocess_models) + (
```

ethicml/evaluators/evaluate_models.py
```diff
@@ -79,49 +80,68 @@ def evaluate_models(datasets: List[Dataset], preprocess_models: List[PreAlgorith
...
-        for pre_process_method in preprocess_models:
-            new_train, new_test = pre_process_method.run(train, test)
-            to_operate_on[pre_process_method.name] = {
-                'train': DataTuple(x=new_train, s=train.s, y=train.y),
-                'test': DataTuple(x=new_test, s=test.s, y=test.y),
-            }
+    seed = 0
+    with tqdm(total=total_experiments) as pbar:
+        for dataset in datasets:
+            for repeat in range(repeats):
```
thomkeh: the repeat is not reflected in the CSV output

thomkeh:
so, I think this is the correct way to compute it:

```python
total_experiments = len(datasets) * repeats * (1 + len(preprocess_models) + (
   (1 + len(preprocess_models)) * len(inprocess_models))
```

ethicml/evaluators/evaluate_models.py
olliethomas: now append repeats to csv if exists

PR #64, opened 2019-03-29
Merge `run()` and `run_threaded()`

thomkeh:
run_threaded() was already basically never called anywhere. Merging these two functions furthermore makes it easier to write Algorithm classes that can only run as subprocesses.


thomkeh:
If you think merging them is going to far, then we should at least make `run_threaded()` private (i.e., rename it to `_run_threaded`).

PR #65, opened 2019-03-29
Don't use Pipenv to run tests

thomkeh:
The dependencies are now all in the `setup.py` and can be installed with `pip install -e .[dev]`.


thomkeh:
Tests run about 20s faster... I had hoped for more.

PR #69, opened 2019-04-16
moved kamiran from pre to in algorithm

olliethomas:
Need to start trying to merge some of my local changes which are well out of sync with master. Expecting this to fail CI but should be a good starting point for getting back to sanity.


olliethomas:
I'm still getting a problem with zemel locally. I think I've hosed my local python :/

PR #101, opened 2019-05-29
Make the threaded run function async

thomkeh:
Algorithms now can implement the function `run_async()` which should run asynchronously. At the moment the `run_async()` implementations are not really asynchronously but that will change with the next PR! Then we can finally run algorithms in parallel.


thomkeh:
I will think about this a bit more.

PR #102, opened 2019-05-30
Run algorithms embarrassingly parallel

thomkeh:
You can try it out with

```python
lr_result, svm_result, maj_result = run_in_parallel([LR(), SVM(), Majority()], data)
```

### DIFF
--- a/ethicml/evaluators/parallelism.py
+++ b/ethicml/evaluators/parallelism.py
@@ -30,7 +30,9 @@ async def arrange_in_parallel(
         assert len(data) == len(algos)
         coroutines = [algo.run_async(train, test) for algo, (train, test) in zip(algos, data)]
     else:
-        assert isinstance(data, tuple)
+        assert len(data) == 2
+        assert isinstance(data[0], DataTuple)
+        assert isinstance(data[1], DataTuple)
         train, test = data
         coroutines = [algo.run_async(train, test) for algo in algos]
     return await asyncio.gather(*coroutines)
--- a/ethicml/evaluators/parallelism.py
+++ b/ethicml/evaluators/parallelism.py
@@ -2,7 +2,7 @@
 Collection of functions that enable parallelism
 """
 import asyncio
-from typing import List, Tuple, Union, Coroutine, Any
+from typing import List, Tuple, Dict
 
 import pandas as pd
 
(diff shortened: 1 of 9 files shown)


ethicml/evaluators/parallelism.py
```diff
@@ -0,0 +1,99 @@
...
+              same length as the list of algorithms
+    Returns:
+        list of the results
+    """
+    coroutines: List[Coroutine[Any, Any, pd.DataFrame]]
+    if isinstance(data, list):
+        assert len(data) == len(algos)
+        coroutines = [algo.run_async(train, test) for algo, (train, test) in zip(algos, data)]
+    else:
+        assert isinstance(data, tuple)
```
olliethomas:
```suggestion
        assert len(data) == 2
        assert isinstance(data[0], DataTuple)
        assert isinstance(data[1], DataTuple)
```
Would be nice to confirm the input is a tuple of DataTuple as this is pretty close to the user.

tests/run_algorithm_test.py
```diff
@@ -24,12 +26,41 @@ def get_train_test() -> Tuple[DataTuple, DataTuple]:
...
+    assert count_true(result[1].values == -1) == 199
+    assert count_true(result[2].values == 1) == 0
+    assert count_true(result[2].values == -1) == 400
+
+
+def test_run_parallel_multi_data():
+    data0 = get_train_test()
+    data1 = get_train_test()
+    data2 = get_train_test()
+    result = run_in_parallel([LR(), SVM(), Majority()], [data0, data1, data2])
```
olliethomas: It would be really nice if there were a 3rd case, where 
`result = run_in_parallel(model(), [data1, data2, data3])`
would apply the one model to multiple datasets. The use case would be where we have created multiple variations on a dataset and we want to run one model over it (as we did recently)

tests/run_algorithm_test.py
```diff
@@ -24,12 +26,41 @@ def get_train_test() -> Tuple[DataTuple, DataTuple]:
...
 
+def test_run_parallel():
+    data = get_train_test()
+    result = run_in_parallel([LR(), SVM(), Majority()], data)
+    assert count_true(result[0].values == 1) == 211
+    assert count_true(result[0].values == -1) == 189
+    assert count_true(result[1].values == 1) == 201
+    assert count_true(result[1].values == -1) == 199
+    assert count_true(result[2].values == 1) == 0
+    assert count_true(result[2].values == -1) == 400
```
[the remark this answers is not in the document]
thomkeh: yes, they're copied from there

PR #105, opened 2019-06-07
added tv metric

olliethomas:
this needs tidying up a bit. Need to think a bit about the best way to deal with discrete predictions. (Or if they need to be handled at all)


olliethomas:
image

PR #110, opened 2019-06-17
Put the classes for logistic regression in one file

thomkeh:
Maybe not worth doing...


thomkeh:
We have too many files... Oliver, you can revert this change if you don't like it.

PR #122, opened 2019-07-01
made the datatuple and testtuple classes iterable

olliethomas:


thomkeh:
I think it's a good idea!

PR #145, opened 2019-08-07
Fair grid search

thomkeh:
I needed to re-organize the code quite a bit because I needed to know which scores belonged to the same run.

I also made use of a NamedTuple. Could have been done without it though.


thomkeh:
I guess I still need a test for the new `get_params_for_top_k()`.

PR #149, opened 2019-08-10
Sort the functions in the numpy type def file alphabetically

thomkeh:


thomkeh:
Beutel and VFAE are randomly failing?

thomkeh:
Ah, might be because of new pytorch version

PR #153, opened 2019-08-15
Add Zafar algorithm

thomkeh:


thomkeh:
@olliethomas do we want to add a test for this? I'm afraid it will be slow.

olliethomas:
sure, I'll do one now

thomkeh:
It would be a bit more efficient if only download the code once for the 3 Zafar algorithms and remove it once. Like I did with fair GPyTorch:

```python
@pytest.fixture(scope="module")
def gpyt_models():
    # the gpyt models are created together because they share a git repository
    gpyt = GPyT(epochs=2, length_scale=2.4)
    gpyt_dem_par = GPyTDemPar(epochs=2, length_scale=0.05)
     gpyt_eq_odds = GPyTEqOdds(epochs=2, tpr=1.0, length_scale=0.05)
    yield (gpyt, gpyt_dem_par, gpyt_eq_odds)
    gpyt.remove()  # only has to be called from one of the models because they share a directory
```

The problem is maybe that this not an entirely realistic test.

olliethomas:
always run the check_all script before pushing, Oliver......

olliethomas:
I agree, I know it will make the test run a bit faster.... Maybe we should have it for the tests. But, In day to day use, downloading all three separately is far more realistic.

One thing that I wasn't sure on is how to go about running the cross validator on the zafar model (in a pretty way)

thomkeh:
> One thing that I wasn't sure on is how to go about running the cross validator on the zafar model (in a pretty way)

What's the problem with

olliethomas:
I think I have an idea

thomkeh:
Can this be merged, @olliethomas ? Or do you want to change the tests in some way?

olliethomas:
I'm fine with it

PR #164, opened 2019-08-23
Make PyTorch a dev requirement only

thomkeh:
Having a specific torch version in the requirements only causes problems it seems.

To make matters worse, it seems that on Windows, you can't install pytorch from pypi at all! You *always* have to install it via a download URL like this
```shell
pip install https://download.pytorch.org/whl/cpu/torch-1.0.1-cp36-cp36m-win_amd64.whl
```
I don't think we can cover all these special cases in our `setup.py`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # EthicML
 
-EthicML exists to combat the problems we've found with off-the-shelf fairness comparison packages.
+EthicML exists to solve the problems we've found with off-the-shelf fairness comparison packages.
 
 These other packages are useful, but given that we primarily do research,
 a lot of the work we do doesn't fit into some nice box.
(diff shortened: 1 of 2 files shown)


thomkeh:
We should probably add something to the README like

> As there are many different versions of PyTorch, it is not possible for us to require the version that fits your case the best. Please install PyTorch manually.

And maybe in all files where we import pytorch, we do this:

```python
try:
    import torch
except ImportError:
    raise RuntimeError("Please install PyTorch >= 1.1.0 to use this feature.")
```

thomkeh:
The updated README can be found here: https://github.com/predictive-analytics-lab/EthicML/blob/torch-as-dev/README.md

PR #168, opened 2019-08-28
 Give option of error crosses instead of boxes

thomkeh:


thomkeh:
Currently looks like this:

Screenshot 2019-08-28 at 22 45 41

PR #191, opened 2019-09-12
Check pylint separately

thomkeh:
Removed the import check from pylint, but added lots of other checks that were disabled.


olliethomas:
If we have a new package that we added to the setup.py, would we then add to pylint.rc as well?

olliethomas:
also, the linting score seems really high. 9.94/10?

thomkeh:
> If we have a new package that we added to the setup.py, would we then add to pylint.rc as well?

Yes.

> also, the linting score seems really high. 9.94/10?

I'm pretty sure I always had this kind of score locally but on Azure we didn't have it that high because pylint thought that some function from torch wasn't callable and that another thing wasn't assignable.

PR #201, opened 2019-09-22
Add a ton of function annotations

thomkeh:
I used this tool to generate most of them: https://github.com/dropbox/pyannotate

### DIFF
--- a/check_all.py
+++ b/check_all.py
@@ -35,7 +35,12 @@ print(MYPY_RESULTS[0], end='')
 print(MYPY_RESULTS[1], end='')
 print("Exit code of mypy: {}".format(MYPY_RESULTS[2]))
 
-# black
-print("############  black ##############")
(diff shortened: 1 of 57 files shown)


check_all.py
```diff
@@ -35,7 +35,12 @@
 print(MYPY_RESULTS[1], end='')
 print("Exit code of mypy: {}".format(MYPY_RESULTS[2]))
 
-# black
-print("############  black ##############")
```
olliethomas: The reason for running on ./ is because black exits and ends the script. The formatting on test/ will never be executed. 

Ideally black would be run before the linter as there wouldn’t be formatting errors, but black has to be run last

thomkeh: Oh, I see. I changed it because the way it was, black was formatting files that I had in that directory (unstaged files) that I didn't want formatted, e.g. code from Zafar

thomkeh: I'll change it back

PR #203, opened 2019-09-25
black formatted

olliethomas:
move from pyproject to specific black config


olliethomas:
@thomkeh I think I've requested the pre-commit hook to only check ethicml and tests, but I'm not entirely sure it's correct. Could you have a look when you get time and see if there'a glaring error?

.black-config.toml
```diff
@@ -0,0 +1,27 @@
+[tool.black]
+line-length = 88
```
thomkeh: new line length?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
+-   repo: https://github.com/ambv/black
+    rev: stable
+    hooks:
+    - id: black
+      language_version: python3.6
+      args: [--config=.black-config.toml]
```
thomkeh: Don't you need our other arguments like `-S` for leaving the quotes alone and `-l 100` for setting the line length?

setup.py
```diff
@@ -46,18 +46,19 @@
...
-        'dev': [
+        "dev": [
             # "black",
             # "mypy >= 0.720",
             # "pylint >= 2.0",
             "pytest >= 3.3.2",
             "pytest-cov >= 2.6.0",
             "torch >= 1.1.0, <= 1.1.0.post2",
+            "pre-commit",
         ]
```
thomkeh: I'm starting to think we should have it actually like this:

```python
        'ci': [
            "pytest >= 3.3.2",
            "pytest-cov >= 2.6.0",
            "torch >= 1.1.0, <= 1.1.0.post2",
        ],
        'dev': [
            "black",
            "mypy >= 0.720",
            "pylint >= 2.0",
            "pytest >= 3.3.2",
            "pytest-cov >= 2.6.0",
            "pre-commit",
        ]
```

and then on the CI we install ethicml with `pip install .[ci]` and locally you can install it with `pip install ethicml[dev]` if you want.

thomkeh:
see inline comments

.pre-commit-config.yaml
olliethomas: I realised that we don't really need the full blown config file, we can just use the args. That said, I still can't get it to not fix the setup.py 🤷‍♂

setup.py
olliethomas: torch just in the ci?

.pre-commit-config.yaml
thomkeh: Ah sorry my comment was stupid. This is of course handled in the .toml file. I'm fine with keeping the config file for black. After all we also have config files for the other linters.

setup.py
thomkeh: Yeah, the way I'm thinking about it is that `dev` is for the human developer who might be on Windows where `pip install torch` doesn't work.

PR #228, opened 2019-12-13
Pass args object directly to load_data_from_flags()

thomkeh:
Getting rid of 38 lines! While making the code more type safe!


thomkeh:
I'm running into the following problem: for the BeutelArgs argparser, the paths are `required=True`, but we don't set them in algorithm class (we don't *need* to set them). How to solve this?

thomkeh:
I reduced the scope of the PR.

PR #244, opened 2020-01-10
Add a decorator for methods implementing an interface

thomkeh:
`pydocstyle` is configured to ignore missing docstrings on such functions.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -103,4 +103,4 @@ jobs:
         pip install pydocstyle
     - name: Check docstrings with pydocstyle
       run: |
-        pydocstyle --convention=google ethicml
+        pydocstyle --convention=google --ignore-decorator=implements ethicml
--- a/check_all.py
+++ b/check_all.py
@@ -13,7 +13,9 @@ print('')
 
 # pydocstyle
 print("############### pydocstyle #################")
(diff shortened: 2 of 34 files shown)


thomkeh:
For

```python
class Accuracy(Metric):
    @implements(Metric)
    def score(self, ...): ...
```

I wanted to implement something in mypy that checks that `Metric` is indeed a superclass of `Accuracy` and that `Metric` does have a function called `score` (whether the signature of `Accuracy.score` matches the one of `Metric.score` is checked anyway already). But it's not possible with mypy.

We could instead have a check in `implements` that does something like:

```python
class implements:
    def __init__(self, base):
        self.base = base

    def __call__(self, func):
        assert hasattr(self.base, func.__name__)  # run-time check
        return func
```

but that would of course introduce some run-time overhead.

olliethomas:
I think the overhead would be minimal tbh

olliethomas:
I like it. One suggestion/idea/discussion point though, in (for example) FairGPyT, the `run_async` function is annotated with `implements(InstalledModel)`. This makes sense as GPyT extends `InstalledModel` and the function is indeed required to be implemented in a `InstalledModel`. But `run_async` isn't defined in `InstalledModel`, the function implemented is declared in `InAlgorithmAsync`. Would it be useful if `run_async` were decorated with `implements(InAlgorithmAsync)`?

PR #245, opened 2020-01-13
Make evaluate_models_parallel() an async function

thomkeh:
And also support pre-processing algorithms in `evaluate_models_parallel()`.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -103,4 +103,4 @@ jobs:
         pip install pydocstyle
     - name: Check docstrings with pydocstyle
       run: |
-        pydocstyle --convention=google --ignore-decorator=implements|overload ethicml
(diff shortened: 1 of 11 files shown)


olliethomas:
The use of generics is 🔥🔥🔥

thomkeh:
The main goal of this was: get rid of the `tornado == 4.5.3` in `setup.py`.

For that, we needed an equivalent of `evaluate_models()` that was `async`. This place has now been taken by `evaluate_models_async()` (formerly known as `evaluate_models_parallel()`). `evaluate_models_async()` is fully compatible with `evaluate_models()` (except for some ordering issues in the results dataframe) and is `async`.

This means that when you need to run an `async`-only algorithm (currently that's only _FairGPyT_, _Kamishima_ and _Zafar_) in a Jupyter notebook, then you should directly use `await`:

```python
results = await evaluate_models_async([Adult()], [], [ZafarAccuracy()])
```

and everything should just work.

PR #246, opened 2020-01-14
Make only torch models (and Agarwal) threaded

### LINKED ISSUE #229: Idea for simplifying algorithms: have each algorithm either run in its own process or not. Don't allow both
With this change, lots of code can be removed and it is then way more straight-forward to add new algorithms.

Now we only have to decide which algorithms should run in their own process and which shouldn't.

- Algorithms using PyTorch: I think they should be in their own process. If you run PyTorch and then keep the process alive, it never really cleans up after itself and often still hogs the GPU.
- Algorithms using Sklearn: I don't see much reason for them to be their own process. Let's run those in the main process.
- "Installed algorithms" that are cloned from GitHub: these just have to run in their own process because they might use a different version of Python

thomkeh:
This has two main reasons:

1. It simplifies the code a lot. Algorithms are now either just threaded or just non-threaded (i.e. they run in the main context of the process). This will hopefully make it easier to implement new algorithms.
2. It is really only models that run on the GPU that benefit from running in their own thread. The problem with these models is that they can often hog the GPU if the process is not properly shut down. This is especially a problem when you run algorithms in a notebook. For the sklearn models this is not really an issue and running them explicitly parallel also doesn't have much of an effect, because they naturally use as many CPUs as they can.

closes #229


thomkeh:
(This removed about 500 lines of code 🎉 )

thomkeh:
The algorithms fall now into the following classes:

- `PreAlgorithm`:
  - Upsampler
- `PreAlgorithmAsync`:
  - Beutel
  - VFAE
  - Zemel
- `InAlgorithm`:
  - Kamiran
  - LR
  - LRProb
  - LRCV
  - MLP
  - SVM
- `InAlgorithmAsync`:
  - Agarwal

PR #249, opened 2020-01-20
Add len operator to dataset classes.

MylesBartlett:


thomkeh:
Just an easy to fix syntax error (wrong indent of docstring)

thomkeh:
Can you also add it as an abstract function to the base class?

thomkeh:
Did you not see that the diff was *enormous*...

PR #258, opened 2020-01-22
Slight improvement that came to me in a dream

thomkeh:
What mypy really was complaining about there in celeba.py wasn't the fact that it wasn't strings, it was the fact that it wasn't a list. So the `map(str, ...)` wasn't actually needed.


olliethomas:
you have great dreams

PR #260, opened 2020-01-22
Introduce a minimal Prediction class

thomkeh:
The prediction class just very simply wraps a pandas Series at the moment. This is just the first step. I wanted to get this merged in two steps because otherwise it gets too complicated to review and so on.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -80,7 +80,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install mypy
-        pip install data-science-types==0.2.2
+        pip install git+https://github.com/predictive-analytics-lab/data-science-types
         pip install typed-argument-parser==1.4
     - name: Type check with mypy
       run: |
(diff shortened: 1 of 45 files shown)


thomkeh:
The prediction class looks like this at the moment:

```python
@dataclass(frozen=True)
class Prediction:
    """Prediction of an algorithm."""

    hard: pd.Series  # hard predictions (e.g. 0 and 1)
```

but I have big plans on how to extend this.

thomkeh:
Test failure is because of missing functions for pd.Series. I address this [here](https://github.com/predictive-analytics-lab/data-science-types/pull/46).

thomkeh:
Maybe do it like this:

```python
class HardPrediction:
    def __init__(self, hard: pd.Series):
        self._hard = hard
    @property
    def hard(self):
        return self._hard

class SoftPrediction(HardPrediction):
    def __init__(self, soft: pd.Series):
        self._soft = soft
        self._hard = heavyside(self._soft)  # soft is automatically converted to hard if desired
    @property
    def soft(self):
        return self._soft
```

PR #269, opened 2020-02-03
Add Calders massaging method

thomkeh:


olliethomas:
It looks as though there’s a lot of overlap with the Upsampler. Maybe we should merge the 2, especially as they are introduced in the same paper.

thomkeh:
It's actually not that much overlap, I think. Sure, both use very similar pandas functionality but the code is only similar and not the same.

You probably could extract a good part of the code into 2-3 functions, but that would be a lot of work...

thomkeh:
I opened an issue: #270

PR #285, opened 2020-02-19
Add CelebA as a proper EthicML dataset

thomkeh:


ethicml/data/health.py
```diff
@@ -14,7 +14,7 @@ def name(self) -> str:
     @property
     def filename(self) -> str:
         """Getter for filename."""
-        return "health.csv"
+        return "health.csv.zip"
```
olliethomas: ???

olliethomas:
```suggestion
        return "health.csv"
```

thomkeh:
You don't like the use of zipped CSV files?

olliethomas:
TBH, I thought that you'd run find and replace on .csv to .csv.zip. I didn't realise that all csvs had been zipped

Issue #125, opened 2019-07-03
Limit API depth


thomkeh:
Can we agree that this is the public API:

```python
algorithms
algorithms.inprocess
algorithms.postprocess
algorithms.preprocess
data
evaluators
metrics
preprocessing
utility
visualisation
```

And that that you never have to go deeper in order to import something?

Examples:

```python
# NOT okay
from ethicml.preprocessing.train_test_split import train_test_split

# okay
from ethicml.preprocessing import train_test_split
```

thomkeh:
We could also maybe shorten `ethicml.algorithms.preprocess` to `ethicml.algos.pre` and so on.

thomkeh:
Done in #128

PR #287, opened 2020-02-20
Fix imnports in algorithms/__init__.py.

MylesBartlett:


ethicml/algorithms/__init__.py
```diff
@@ -1,3 +1,4 @@
 """Definitions of algorithms."""
-from . import inprocess, preprocess
 from .algorithm_base import *
+from .inprocess import *
+from .preprocess import *
```
thomkeh: We decided to not do that. See #125

PR #296, opened 2020-02-24
Restructure celeba

thomkeh:


olliethomas:
I'm reluctant to approve a change to the .pylintrc file

olliethomas:
I think there is an argument to be made that 

```
if X:
    return foo
else:
    return bar
```

is more readable than

```
if X:
    return foo
return bar
```

But I think that having this enforced encourages the developer to think more

thomkeh:
I expected you to say that.

As I see it, the reason why pylint doesn't like this:
```
if something:
    return 0
else:
    return 1
```
is that you might _forget_ to return in one of the if-else branches (this is especially true when the branches are long and intricate). But in our case, mypy already checks that we always return. And even makes sure that we return the correct type. So, I think the pylint check is superfluous.

olliethomas:
Provided that mypy remains strict, I am persuaded.

PR #305, opened 2020-03-13
Use numpy files instead of feather files, second attempt

thomkeh:
This only affects inprocess algorithms right now.

### DIFF
--- a/tests/saving_data_test.py
+++ b/tests/saving_data_test.py
@@ -18,15 +18,15 @@ def test_simple_saving() -> None:
     tests that a DataTuple can be saved
     """
     data_tuple = DataTuple(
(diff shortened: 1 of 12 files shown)


tests/saving_data_test.py
```diff
@@ -18,15 +18,15 @@ def test_simple_saving() -> None:
     tests that a DataTuple can be saved
```
MylesBartlett:
```suggestion
    Tests that a DataTuple can be saved.
```

thomkeh: I'm not committing that because then I'd have to wait for another 6 minutes.

thomkeh:
Thank you for your review!

PR #310, opened 2020-03-23
Add new depenendence measures

thomkeh:
I also turned off `per_sens_attribute` for NMI. I don't think it makes sense to apply NMI to y and yhat. It's better to use accuracy for that. NMI only really shines when we don't know how the categories of two variables correspond to each other.

I also did some experiments to compare the values that the new metrics give. If we just compute DI (i.e. prob pos ratio) on the Adult dataset for s and y (pretending y is our prediction), then we get **0.364**. Measuring the dependence between s and y with the new metrics we get:

| measure      | value  |
|:--------------|--------|
| NMI          | 0.0432 |
| Yanovich     | 0.0396 |
| Renyi | 0.216  |

So, the one that gave the most reasonable number was "Renyi" from [this](http://proceedings.mlr.press/v97/mary19a/mary19a-supp.pdf) paper.

What these measures all share is that they are 0 if s and yhat are independent and 1 if s and y are the same (or the same when taking the inverse or something like that).

See the tests for more examples.


ethicml/metrics/dependence_measures.py
```diff
@@ -0,0 +1,144 @@
+"""For assessing the Mutual Information between s and yhat."""
+from typing import ClassVar
```
olliethomas: I didn't know about this type

ethicml/metrics/dependence_measures.py
```diff
@@ -0,0 +1,144 @@
...
+
+    _base_name: ClassVar[str] = "Witsenhausen"
+
+    @implements(Metric)
+    def score(self, prediction: Prediction, actual: DataTuple) -> float:
+        base_values = actual.y if self.base == "y" else actual.s
+        return _witsenhausen(base_values.to_numpy().ravel(), prediction.hard.to_numpy().ravel())
+
+
+def _witsenhausen(x: np.ndarray, y: np.ndarray) -> float:
```
olliethomas: nitpicking, but maybe this should be a static method of the Witsenhausen class?

olliethomas: same with yanovich

ethicml/metrics/dependence_measures.py
thomkeh: I also only discovered it recently

thomkeh:
Here are some more numbers:

image

This is evaluating the dependence of s and the label y on the Adult dataset with different mixing factors.

As you can see Yanovich and NMI are nearly the same. Also, the Renyi correlation very closely tracks DP diff. I'm surprised by this. Renyi correlation is this complicated algorithm based on singular value decomposition, and DP diff is just subtraction.

PR #311, opened 2020-03-24
add notebook describing how to create the toy dataset

olliethomas:
Update the tests to reflect the new dataset. Also, update precommit setting for autoflake.

Can't run installed models locally...... not sure how to proceed with that


.pre-commit-config.yaml
```diff
@@ -3,7 +3,7 @@ repos:
     rev: v1.1
     hooks:
     - id: autoflake
-      args: ['--in-place', '--remove-all-unused-imports', '--remove-unused-variable']
+      args: ['-h', '--in-place', '--remove-all-unused-imports', '--ignore-init-module-imports']
```
thomkeh: do you need the `-h`?

olliethomas: Nope! Good catch

olliethomas:
I've been thinking about this. My hesitation in merging this in is because I don't think it's a very comprehensive way of adding bias to a dataset, what I ideally wanted was a toy dataset with lots of types of unfairness in the data. I now don't think this is a good idea. Instead, I think a better target would be to have a toy dataset that fits the form of the data so that we can use it for testing and add further synthetic datasets in the future that display specific types of bias/unfairness.

So the datasets may end upsomething like this
- toy (for quick testing of models)
- synthetic_selection_bias
- synthetic_proxy_bias
- synthetic_unfair_labeler

thomkeh:
Yes that sounds good. I do think we need a small toy dataset just for running tests which doesn't have to be especially realistic.

olliethomas:
> Yes that sounds good. I do think we need a small toy dataset just for running tests which doesn't have to be especially realistic.

Maybe I should make the toy dataset 200 samples instead of 2000

olliethomas:
ok - made a load of changes, but I can't run the Zafar model and it doesn't seem like the CI can either. So still a bit unsure on how to progress

thomkeh:
Maybe we should take this opportunity to update pytorch in the CI to version 1.5

thomkeh:
The test failure is weird:

```
E FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/EthicML/EthicML/zafar/fair-classification/.venv/bin/python': '/home/runner/work/EthicML/EthicML/zafar/fair-classification/.venv/bin/python' 


```

olliethomas:
I figured that this was a clue`/home/runner/work/EthicML/EthicML` but I'm not sure where it's being set

olliethomas:
I'm also not sure how this PR *adds* more lines than it deletes. For a start the toy.csv has gone from 2000 line to 400

tests/conftest.py
```diff
@@ -79,3 +89,9 @@ def temp_dir() -> Generator[Path, None, None]:
     """
     with tempfile.TemporaryDirectory() as tmpdir:
         yield Path(tmpdir)
+
+
+def get_id(value):
+    if hasattr(value, "name"):
+        return value.name
+    return value
```
thomkeh, line 97:
```suggestion
    return getattr(value, "name", value)
```

tests/metrics/metric_test.py
```diff
@@ -0,0 +1,50 @@
...
+class MetricTest(NamedTuple):
+    """Define a test for a metric."""
+
+    model: InAlgorithm
+    metric: Metric
+    name: str
+    expected: float
+
+    def __str__(self):
+        return "asdafs"
```
thomkeh, line 22: ?

tests/metrics/monolith_metric_test.py
```diff
@@ -0,0 +1,416 @@
+"""Test that we can get some metrics on predictions"""
```
thomkeh, line 1: maybe a short explanation what "monolith" refers to?

tests/models_test/inprocess_test/models_inprocessing_test.py
```diff
@@ -262,35 +189,51 @@ def test_zafar(zafar_models, toy_train_test: TrainTestPair) -> None:
...
 
     predictions = run_blocking(model.run_async(train, test))
-    assert predictions.hard.values[predictions.hard.values == 1].shape[0] == 215
-    assert predictions.hard.values[predictions.hard.values == -1].shape[0] == 185
+    expected_num_pos = 242
+    assert predictions.hard.values[predictions.hard.values == 1].shape[0] == expected_num_pos
+    assert (
+        predictions.hard.values[predictions.hard.values == 0].shape[0]
+        == len(predictions) - expected_num_pos
+    )
```
thomkeh, line 236: in 2 lines instead of 4 lines:
```suggestion
    num_neg = predictions.hard.values[predictions.hard.values == 0].shape[0]
    assert num_neg == len(predictions) - expected_num_pos
```

tests/metrics/metric_test.py
olliethomas: 😓

olliethomas:
I've tried cloning Zafar and installing it manually, but I get the same error :(

ethicml/algorithms/postprocess/hardt.py
```diff
@@ -144,7 +144,7 @@ def _predict(
         test_preds_numpy: np.ndarray = test_predictions.hard.to_numpy()
 
         # Randomly flip labels according to the probabilities in model_params
-        self_fair_pred = test_preds_numpy[test.s[test.s.columns[0]].to_numpy() == 0].copy()
+        self_fair_pred = test_preds_numpy[mask_s1].copy()
```
thomkeh, line 147: btw, how did you come across this?

olliethomas: I have no recollection of changing this. If I had to guess, mypy not liking it?

thomkeh: oh wait, it's actually wrong. it should be `mask_s0` instead of `mask_s1`, right?

thomkeh: then why did the tests pass...?

olliethomas: wow... that was a silly mistake... Sorry about that.
I guess there's 2 things happening.
1) Some labels in a group are still being flipped, they're just the other group
2) The tests probably didn't pass, but this was attributed to the data changing as opposed to the model

olliethomas: so, I've just reverted this, and it doesn't run ...

```
self_fair_pred[p2n_indices] = self._unfavorable_label
IndexError: index 39 is out of bounds for axis 0 with size 38
```

PR #316, opened 2020-04-20
Add discrete feature groups to COMPAS dataset

thomkeh:


ethicml/data/util.py
```diff
@@ -92,3 +91,11 @@ def _first_segment(feature_name: str) -> str:
         start_idx += len_group
 
     return feature_slices
+
+
+def flatten_dict(dictionary: Mapping[str, List[str]]) -> List[str]:
+    """Flatten a dictionary of lists by joining all lists to one big list."""
+    flattened: List[str] = []
+    for group in dictionary.values():
```
MylesBartlett, line 99:
```suggestion
    return list(dictionary.values())
```

thomkeh: this doesn't flatten the lists

thomkeh: there might be a better way to do this, but it's not that easy

MylesBartlett: Ah yes, you're right

MylesBartlett:
```suggestion
    return [feat for group in dictionary.values() for feat in group]
```

thomkeh: done

PR #329, opened 2020-05-15
Turn the Results class into a NewType(DataFrame)

thomkeh:
It was getting annoying to always do `results.data` to access the
underlying dataframe. Now that Results is just a NewType wrapper around
DataFrame, you don't have to do this anymore. The price we pay for this
is that all the class methods have to be turned into normal functions
which is a bit ugly, but the "enduser" will likely never deal with them.
(The only place in the code where this comes up is in evaluate_models).

### DIFF
--- a/ethicml/evaluators/cross_validator.py
+++ b/ethicml/evaluators/cross_validator.py
@@ -2,7 +2,7 @@
 from collections import defaultdict
 from itertools import product
 from statistics import mean
-from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Type
+from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Type, Sequence, Mapping
 
 from ethicml.algorithms.inprocess.in_algorithm import InAlgorithm
 from ethicml.metrics import AbsCV, Accuracy, Metric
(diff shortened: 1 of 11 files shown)


ethicml/evaluators/cross_validator.py
```diff
@@ -131,7 +131,7 @@ class CrossValidator:
     def __init__(
         self,
         model: Type[InAlgorithm],
-        hyperparams: Dict[str, List[Any]],
+        hyperparams: Mapping[str, Sequence[Any]],
```
thomkeh, line 134: This solved a mypy complaint from one of the tests.

thomkeh:
More context on `NewType`: https://mypy.readthedocs.io/en/stable/more_types.html#newtypes

ethicml/utility/data_structures.py
```diff
@@ -322,100 +332,91 @@ class TrainValPair(NamedTuple):
     validation: DataTuple
 
 
-class Results:
-    """Container for results from `evaluate_models`."""
+Results = NewType("Results", pd.DataFrame)  # Container for results from `evaluate_models`
```
thomkeh, line 335: This is the new definition of `Results`.

PR #331, opened 2020-05-28
Fwd loss

olliethomas:


olliethomas:
Could potentially add a model that uses the vision datasets

ethicml/utility/lookup.py
```diff
@@ -0,0 +1,25 @@
...
+from typing import Dict
+
+from ethicml.data import Dataset, adult, compas, credit, german, nonbinary_toy, sqf, toy
+
+__all__ = ["get_dataset_obj_by_name"]
+
+
+def get_dataset_obj_by_name(name: str) -> Dataset:
+    """Given a dataset name, get the corresponding dataset object."""
+    lookup: Dict[str, Dataset] = {
```
thomkeh, line 12:
```suggestion
    lookup: Dict[str, Callable[[], Dataset] = {
```

We could also make it so that we just look up the function and only then call it to get the dataset object.

olliethomas: I'm not sure as this isn't something that's used all that often. In fact, given the recent changes to Dataset, I'm not sure if this is still needed, but it probably falls out of the scope of this PR to check that

thomkeh: yeah it was probably a micro optimization

thomkeh:
maybe mention "DRO" (distributionally robust optimization) somewhere?

ethicml/implementations/fwd_modules/fwd_loss.py
```diff
@@ -0,0 +1,25 @@
...
+
+from torch import Tensor, nn
+from torch.nn.modules.loss import NLLLoss, _Loss
+
+from ethicml.common import implements
+
+__all__ = ["FWDLoss"]
+
+
+class FWDLoss(nn.Module):
```
MylesBartlett, line 12:
```suggestion
class DROLoss(nn.Module):
```

ethicml/implementations/fwd_modules/fwd_loss.py
```diff
@@ -0,0 +1,25 @@
+"""FWD Loss."""
+from typing import Optional
+
+from torch import Tensor, nn
+from torch.nn.modules.loss import NLLLoss, _Loss
+
+from ethicml.common import implements
+
+__all__ = ["FWDLoss"]
```
MylesBartlett, line 9:
```suggestion
__all__ = ["DROLoss"]
```

ethicml/implementations/fwd_tabular.py
```diff
@@ -0,0 +1,92 @@
...
+from torch.utils.data import DataLoader
+
+from ethicml.implementations.beutel import set_seed
+from ethicml.implementations.fwd_modules.fwd_classifier import FWDClassifier
+from ethicml.implementations.pytorch_common import CustomDataset, TestDataset
+from ethicml.implementations.utils import InAlgoArgs, load_data_from_flags
+from ethicml.utility import DataTuple, SoftPrediction, TestTuple
+
+
+class FwdArgs(InAlgoArgs):
```
MylesBartlett, line 18:
```suggestion
class DroArgs(InAlgoArgs):
```

ethicml/implementations/fwd_modules/fwd_loss.py
```diff
@@ -0,0 +1,25 @@
...
+
+class FWDLoss(nn.Module):
+    """Fairness Without Demographics Loss."""
+
+    def __init__(self, loss_module: Optional[_Loss] = None, eta: float = 0.5):
+        """Set up the loss, set which loss you want to optimize and the eta to offset by."""
+        super().__init__()
+        if loss_module is None:
+            loss_module = NLLLoss()
+        self.loss = loss_module
```
MylesBartlett, line 20:
```suggestion
        else:
            assert loss_module.reduction == "none"
        self.loss = loss_module
```

PR #345, opened 2020-06-16
Add option to turn off diffs and ratios

thomkeh:
This would be useful when the sensitive attribute is very high-dimensional, because in that case there would be a very large number of diffs and ratios (they're computed pair-wise).


olliethomas:
maybe the pairwise diffs and ratios should be pulled out to a separate function. Just thinking there might be a situation where you are experimenting with high S-dimensional data and you at one point will want the diffs (or mean of the diffs) and a 2-step approach might be nice.

thomkeh:
Right. The only problem with this is the order of the columns. Let me think how best to do it...

PR #348, opened 2020-07-28
update pillow based on security vulnerability email

olliethomas:


olliethomas:
I have no idea why this test failed.... it doesn't seem like a good thing that it did

olliethomas:
even better, the test *passes* locally

olliethomas:
So I updated all my dependencies, the only difference between locally and the GH action is the python version (3.7 local vs 3.6 GH). 
The same test fails, but with `assertion error 41 == 43` locally instead of `assertion error 40 == 43` on GH.

Maybe the zemel method needs reimplementing. We're not measuring something that we've written, this came from AIF360, and it's given us a ton of problems

PR #351, opened 2020-08-12
Make tabular data available as a list

olliethomas:
Instead of having to know which datasets are available, just import the list of available datasets. 

I want to do this for the image datasets too, but they are not as uniform.


olliethomas:
only requesting the review as the lookup function has changed (slightly), and I want to let you know in case you're using it somewhere

ethicml/data/__init__.py
```diff
@@ -13,3 +13,14 @@
...
+available_tabular = [
+    adult.__name__,
+    compas.__name__,
+    credit.__name__,
+    crime.__name__,
+    german.__name__,
+    health.__name__,
+    sqf.__name__,
+    toy.__name__,
+]
```
thomkeh, line 26: Do you wanna add a quick test for this?

ethicml/utility/lookup.py
```diff
@@ -2,24 +2,37 @@
 
 from typing import Callable, Dict
 
-from ethicml.data import Dataset, adult, compas, credit, german, nonbinary_toy, sqf, toy
+from ethicml.data import (
```
thomkeh, line 5: actually, I don't completely love this circular import. `utility` should depend on no other module. `data` should only depend on `utility`. all the other modules may depend on both `data` and `utility`.

thomkeh:
Hmm, why are the tests slower? A fluke?

olliethomas:
Not a fluke - I'm running the tests for scaling over all datasets
TBH, now that I know that it works, I could maybe drop the long-running ones

tests/loading_data_test.py
```diff
@@ -240,53 +191,14 @@ def test_load_adult_race():
...
-    """test load compas race"""
-    data: DataTuple = compas("Race").load()
-    assert (6167, 400) == data.x.shape
-    assert (6167, 1) == data.s.shape
-    assert (6167, 1) == data.y.shape
-
-
 def test_load_adult_race_sex():
-    """test load adult race sex"""
+    """Test load of split combinatin that doesn't exist yet."""
```
thomkeh, line 195: ah, this isn't a test of a split combination that doesn't exist. the problem here is something else: it's a multi s setup that's very complicated, because it's combining a binary attribute (sex) with a one-hot encoded attribute (race) which requires some complicated code to make it work. I hope that one day, this doesn't throw an assertion error.

olliethomas: aha, I'll revert the comment. Multi S is tricky. How do big projects discuss this sort of thing? With a proposal?

thomkeh: yes! write a proposal!

PR #356, opened 2020-08-17
Random classifier

olliethomas:


thomkeh:
I don't think the seeding will work like that, because the seeding is global and you don't know what will happen between the call to init and the call to run().

I know I'm the only one who likes it, but I think you should use numpy's RandomState here.

```python
self.random = np.RandomState(seed=<whatever>)
# ...
self.random.choice(...)
```

olliethomas:
Good point

olliethomas:
`RandomState` wasn't being consistent. I don't know why

olliethomas:
> `RandomState` wasn't being consistent. I don't know why

Because I was doing it wrong

ethicml/algorithms/inprocess/blind.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self, seed: int = 888) -> None:
+        """Init Majority."""
+        super().__init__(name="Blind", is_fairness_algo=False)
+        self.seed = seed
+
+    @implements(InAlgorithm)
+    def run(self, train: DataTuple, test: TestTuple) -> Prediction:
+        train_y_vals = train.y.drop_duplicates()
+
+        self.random = np.random.RandomState(self.seed)
```
thomkeh, line 26: ah, if you do it like this, you don't need the `self.`.

my idea was to put this line into the `__init__` (then you would need to assign the random state to `self`).

not sure which way is better. do you want subsequent runs to have the same random state? I don't know.

thomkeh:
"Blind" is a cool name ;)

ethicml/algorithms/inprocess/blind.py
olliethomas: I was thinking about this, because I don't think it's ever really been discussed before. There are 2 sources of randomness, that I can tell - the data, and the model. 

When we run a cross validation, for example, I think that only the data should change. But, if we're running multiple repeats, I'm not sure if we should just change the data, or change the data and the model randomness together. 

I don't think there's a case where we just want randomness in the model... (that I can think of)

thomkeh: > a case where we just want randomness in the model

uncertainty estimates? :D

thomkeh: @MylesBartlett do you have an opinion on this?

thomkeh: Thinking about it more, you're right that it doesn't make sense that subsequent runs have different seeds. So, leave it as it is.

ethicml/algorithms/inprocess/blind.py
```diff
@@ -0,0 +1,32 @@
...
+
+        predictions = Prediction(
+            hard=pd.Series(self.random.choice(train_y_vals.T.to_numpy()[0], test.x.shape[0]))
```
thomkeh, line 29: unneeded `self`.

```suggestion
        random = np.random.RandomState(self.seed)

        predictions = Prediction(
            hard=pd.Series(random.choice(train_y_vals.T.to_numpy()[0], test.x.shape[0]))
```

ethicml/algorithms/inprocess/blind.py
olliethomas: > > a case where we just want randomness in the model
> 
> uncertainty estimates? :D

Yes, but I think that this should probably be handled by the model itself

PR #358, opened 2020-08-17
Re-work how multi S works

### LINKED ISSUE #353: Proposal: Multiple S
### Abstract
We currently only really support one group of binary S values. FairML literature has a large gap in terms of analysis of a more real-world setup: multiple groups of S. This proposal is for multiple S groups, which in turn are exclusive, to be supported.

For example, currently we have `{gender=0, gender=1}`. This has recently been extended to include more than one sensitive group, `{race0-gender0=0, race1-gender0=1, race2-gender0=2, race0-gender1=3, race1-gender1=4, race2-gender1=5}`. 

This proposal is to formalise this process a little.

### Motivation 
Fairness and Equality laws around the world consider multiple sensitive groups. For example you should be fair with regard to race, gender AND religion (amongst others). But how to best represent this? 

### Specification 

The recent approach is to combine all potential groups giving N sensitive values. During training we can potentially OHE the sensitive value, and during evaluation we can then treat the result as a (1v1/1vA) binary problem. Throughout we can use a lookup table to map back from 3 -> race=1, gender=1. This lookup table would exist as part of the created Dataset object.

#### Lookup Table

The lookup table should map between 2 tables. One which is held in DataTupleInstance.s, and another, more readable one

_DataTupleInstance.s_
| Index | S |
|:-----:|:-:|
|     0 | 0 |
|     1 | 4 |
|     2 | 3 |
|     3 | 1 |
|     4 | 3 |
|     5 | 5 |
|     6 | 2 |
| ...| ...|

There should be a mapping between these 2 tables

_readable_
|       | Gender  |         | Race  |       | |
|-------|---------|---------|-------|-------|-----|
| Index | Gender0 | Gender1 | Race0 | Race1 | Race2 |
|     0 |    1    | 0       | 1     | 0     | 0 |
|     1 |    0    | 1       | 0     | 1     | 0 |
|     2 |    0    | 1       | 1     | 0     | 0 |
|     3 |    1    | 0       | 0     | 1     | 0 |
|     4 |    0    | 1       | 1     | 0     | 0 |
|     5 |    0    | 1       | 0     | 0     | 1 |
|     6 |    1    | 0       | 0     | 0     | 1 |
|     ... |   ...    | ...       | ...     | ...     | ... |

Why multi-hierarchy? This will make it easier to look at just Gender, or just Race

#### Evaluation Metrics
If you're woking on intersectional data groups, then we will need this mapping to produce metrics that are readable. The `run_metrics` helper should optionally accept a Dataset object. If it receives this, then the data should be unpacked prior to reporting the values. If a dataset is not provided, then the evaluations should not try to map from a number `15` to a text description `sex=value0,gender=value0,nationality=value15`.

### Rejected Ideas 

Not collapsing the intersectional groups and leaving just the hierarchical 

### Outstanding Problems

How to extend to continuous S? - We don't have any datasets where this is the case, so not a primary concern.

What if, as a user, I don't want to create a lookup mechanism? I just want to get some metrics for the data I put in? - These lookup tables will be provided for all datasets in EthicML. Unless you're loading in your own data directly into a DataTuple, in which case if no Dataset is provided to one of the evaluation helpers, then the values are assumed to belong to one S group.

What if I have an S group which has no value, or a missing value? This will need to be remedied prior to using EthicML. We assume that each sensitive group has exactly one value per row.

thomkeh:
closes #353


ethicml/utility/data_helpers.py
```diff
@@ -1,9 +1,34 @@
...
+    df: pd.DataFrame, new_column_name: Optional[str] = None
+) -> Union[pd.Series, pd.DataFrame]:
+    """Undo one-hot encoding."""
+    # we have to overwrite the column names because `idxmax` uses the column names
+    df.columns = pd.Index(list(range(df.shape[1])))
+    result = df.idxmax(axis="columns")
+    if new_column_name is not None:
+        return result.to_frame(name=new_column_name)
+    else:
+        return result
```
olliethomas, line 34: Do the ellipsis serve some purpose? Or can they be replaced with a `raise NotImplementedError()`?

tests/loading_data_test.py
```diff
@@ -450,3 +451,36 @@ def test_genfaces():
...
+
+def test_expand_s():
+    """Test expanding s."""
+    data = Dataset(
+        name="test",
+        filename_or_path="non-existent",
+        features=[],
+        cont_features=[],
+        sens_attr_spec={
+            "Gender": LabelSpec(["Female", "Male"], multiplier=3),
```
olliethomas, line 464: Is there a good way to determine the multiplier?

olliethomas: Just trying to work out if there's a way to hide this detail from the user. It might not be possible

thomkeh: this can definitely be automated, but I'm hesitant to do so at this point. the biggest problem with the previous solution was too much automation, which made it very inflexible. so I'm a bit scared of automation now.

ethicml/utility/data_helpers.py
thomkeh: you mean the ellipses from the overloads? I think it can also be `pass`. not sure about `NotImplementedError`.

tests/loading_data_test.py
olliethomas: fair enough. I like the idea of 2 levels working together, such as a higher level where things are more automated e.g. `run_suite`, and lower level where things are actually done

olliethomas: but not saying such a thing needs to (or could) exist for this!

ethicml/utility/data_helpers.py
olliethomas: oh - I think I got myself confused. I thought they were saying if called with this function signature do something different. I didn't realise that they were there just for mypy

tests/loading_data_test.py
thomkeh: hm, I guess there could just be a function that covers the most common use cases of this... ok I'll add it

thomkeh: done

ethicml/data/tabular_data/adult.py
```diff
@@ -149,91 +150,38 @@ def adult(
...
-            "native-country_South",
-            "native-country_Taiwan",
-            "native-country_Thailand",
-            "native-country_Trinadad&Tobago",
-            "native-country_United-States",
-            "native-country_Vietnam",
-            "native-country_Yugoslavia",
-        ]
+        sens = "native-country"
+        sens_attr_spec = simple_spec({sens: disc_feature_groups[sens]})
```
thomkeh, line 182: this is the code I was talking about

```suggestion
        sens_attr_spec = simple_spec({(key := "native-country"): disc_feature_groups[key]})
```

PR #365, opened 2020-08-20
add synthetic 3

olliethomas:
also regenerated synthetic 2 as there was an incorrect specification


thomkeh:
What was black doing that you had to turn it off?

olliethomas:
it was putting some of the parameterizations that were too long to 1 line per parameter. Which wasn't very useful as it made the tests harder

PR #367, opened 2020-08-21
Import everything in ethicml into one namespace

thomkeh:
Is there a reason not to do this?

(`ethicml.implementations` and `ethicml.vision` are of course not automatically imported, because they require pytorch.)

### DIFF
--- a/ethicml/__init__.py
+++ b/ethicml/__init__.py
@@ -1,2 +1,9 @@
 """EthicML."""
-from . import algorithms, common, data, evaluators, metrics, preprocessing, utility, visualisation
+from .algorithms import *
+from .common import *
+from .data import *
+from .evaluators import *
+from .metrics import *
+from .preprocessing import *
+from .utility import *
+from .visualisation import *
--- a/ethicml/algorithms/preprocess/__init__.py
+++ b/ethicml/algorithms/preprocess/__init__.py
@@ -1,7 +1,7 @@
 """Pre-process algorithms take the training data and transform it."""
-from .beutel import Beutel
-from .calders import Calders
-from .pre_algorithm import PreAlgorithm, PreAlgorithmAsync
-from .upsampler import Upsampler
-from .vfae import VFAE
-from .zemel import Zemel
+from .beutel import *
+from .calders import *
+from .pre_algorithm import *
+from .upsampler import *
+from .vfae import *
+from .zemel import *
--- a/ethicml/algorithms/preprocess/beutel.py
+++ b/ethicml/algorithms/preprocess/beutel.py
@@ -7,6 +7,8 @@ from ethicml.utility import FairnessType
 from .interface import flag_interface
 from .pre_algorithm import PreAlgorithmAsync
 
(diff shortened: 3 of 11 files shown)


olliethomas:
I don't think there's any major reason to _not_ do this. We don't have name clashes and there aren't so many things you'd want to import that there could really be any confusion between `aml.adult` and `eml.data.adult`. The only thing that would perhaps be useful is something in the docs e.g. when you look up ethicml > data > tabular data > adult, it says that it can be used as `eml.adult`. I don't know if this is possible though

olliethomas (approved):

thomkeh:
That's doable but would require some custom code. I just had a look at how TensorFlow does it. For example, the `argmax` function is accessible through both `tf.argmax` and `tf.math.argmax`. The code looks like this:

```python
@tf_export("math.argmax", "argmax", v1=[])
def argmax_v2(input, axis=None, output_type=dtypes.int64, name=None):
  """Returns the index with the largest value across axes of a tensor.
  In case of identity returns the smallest index.
  For example:
  >>> A = tf.constant([2, 20, 30, 3, 6])
  >>> tf.math.argmax(A)  # A[2] is maximum in tensor A
  <tf.Tensor: shape=(), dtype=int64, numpy=2>
  >>> B = tf.constant([[2, 20, 30, 3, 6], [3, 11, 16, 1, 8],
  ...                  [14, 45, 23, 5, 27]])
  >>> tf.math.argmax(B, 0)
  <tf.Tensor: shape=(5,), dtype=int64, numpy=array([2, 2, 0, 2, 2])>
  >>> tf.math.argmax(B, 1)
  <tf.Tensor: shape=(3,), dtype=int64, numpy=array([2, 2, 1])>
  >>> C = tf.constant([0, 0, 0, 0])
  >>> tf.math.argmax(C) # Returns smallest index in case of ties
  <tf.Tensor: shape=(), dtype=int64, numpy=0>
  Args:
    input: A `Tensor`.
    axis: An integer, the axis to reduce across. Default to 0.
    output_type: An optional output dtype (`tf.int32` or `tf.int64`). Defaults
      to `tf.int64`.
    name: An optional name for the operation.
  Returns:
    A `Tensor` of type `output_type`.
  """
  if axis is None:
    axis = 0
  return gen_math_ops.arg_max(input, axis, name=name, output_type=output_type)
```

So, it seems they have a `@tf.export()` decorator which handles this.

olliethomas:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/util/tf_export.py

olliethomas:
test file https://github.com/tensorflow/tensorflow/blob/f618ab49554dba03c104ef27cea5d893063fd373/tensorflow/python/util/tf_export_test.py

thomkeh:
That seems a dead end. Now I'm thinking their docs generator must do it somewhere: https://github.com/tensorflow/docs/blob/master/tools/tensorflow_docs/api_generator/parser.py

thomkeh:
I'll try to improve the docs in a follow-up PR.

PR #370, opened 2020-08-24
Improve docs

thomkeh:
Looks like this now:

image

previously: https://predictive-analytics-lab.com/EthicML/

I also removed a lot of pointless docstrings for `__init__` methods. These docstrings get appended to the class docstring. So, for example this:
```python
class Accuracy(SklearnMetric):
    """Classification accuracy."""

    def __init__(self, pos_class: Optional[int] = None):
        """Init Accuracy."""
        # ...
```

results in this:
image

So you see, the `__init__` docstring really doesn't add much.


thomkeh:
If you want to suggest better names for the submodules:

- Algorithms base
- Preprocess algorithms
- Inprocess algorithms
- Postprocess algorithms
- Datasets
- Evaluation helpers
- Metrics
- Data preprocessing
- Utils
- Vision data helpers
- Visualisation helpers

I'm all ears!

thomkeh:
Looks like this now:

image

PR #372, opened 2020-08-25
Remove pillow version and fix version number on master

thomkeh:
The pillow dependency was only there because one version of pytorch had a problem with pillow 7.0. But this is not the case anymore, so we're not requiring any specific version of pillow.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,9 +5,9 @@ from setuptools import find_packages, setup
 # We follow Semantic Versioning (https://semver.org/)
 _MAJOR_VERSION = "0"
 _MINOR_VERSION = "1"
-_PATCH_VERSION = "0"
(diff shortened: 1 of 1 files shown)


MylesBartlett (approved):

olliethomas:
Pillow 7.0 (and I assume <7) has a security vulnerability. To make github's dependabot not message persistently we need >=7.1

Example email

olliethomas,
We found a vulnerable dependency in a repository you have security alert access to.

@predictive-analytics-lab	predictive-analytics-lab/nifr
Known moderate severity security vulnerability detected in Pillow < 7.1.0 defined in setup.py.

thomkeh:
Okay, then let's just remove the dependency entirely.

PR #374, opened 2020-08-26
Write tests to simulate the case where no pytorch is available

thomkeh:
This turned out to be a bit tricky. It seems the best way to do this is to have a central point that checks if torch can be imported and which then sets a global variable. Then, in the tests we can manipulate that variable to make the code think that torch is not installed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,7 @@ from typing import Generator
 import pandas as pd
 import pytest
 
+import ethicml
 from ethicml.common import ROOT_PATH
 from ethicml.data import load_data, toy
 from ethicml.preprocessing import train_test_split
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -86,3 +87,18 @@ def temp_dir() -> Generator[Path, None, None]:
...
+    torch_available = ethicml.common.TORCH_AVAILABLE
+    torchvision_available = ethicml.common.TORCHVISION_AVAILABLE
+    ethicml.common.TORCH_AVAILABLE = False
+    ethicml.common.TORCHVISION_AVAILABLE = False
+
+    yield  # run test
+
+    # ====== tear down =======
+    ethicml.common.TORCH_AVAILABLE = torch_available
+    ethicml.common.TORCHVISION_AVAILABLE = torchvision_available
```
thomkeh, line 104:
```suggestion
    torch_available = em.common.TORCH_AVAILABLE
    torchvision_available = em.common.TORCHVISION_AVAILABLE
    em.common.TORCH_AVAILABLE = False
    em.common.TORCHVISION_AVAILABLE = False

    yield  # run test

    # ====== tear down =======
    em.common.TORCH_AVAILABLE = torch_available
    em.common.TORCHVISION_AVAILABLE = torchvision_available
```

PR #377, opened 2020-08-26
Do wildcard import only from modules with __all__ defined

thomkeh:
supersedes #376


thomkeh:
I'll merge this. While it's verbose, it's also pretty safe and shouldn't break anything. We can think of a better solution later.

PR #378, opened 2020-08-26
Try out `import ethicml as em` in run_algorithm_test.py

thomkeh:


thomkeh:
This should, again, be a pretty harmless change.

PR #392, opened 2020-10-23
pass info

olliethomas:
Basically, I want to be able to pass additional information from an Async Algo to the main calling script. 
There's already an `info` parameter, so now just write this value to json, then load if exists.

### DIFF
--- a/ethicml/utility/data_structures.py
+++ b/ethicml/utility/data_structures.py
@@ -1,4 +1,5 @@
 """Data structures that are used throughout the code."""
+import json
 from pathlib import Path
 from typing import (
     Callable,
@@ -47,6 +48,7 @@ class TestTuple:
     """A tuple of dataframes for the features and the sensitive attribute."""
 
     def __init__(self, x: pd.DataFrame, s: pd.DataFrame, name: Optional[str] = None):
+        """Make a TestTuple."""
         self.__x: pd.DataFrame = x
         self.__s: pd.DataFrame = s
         self.__name: Optional[str] = name
@@ -118,6 +120,7 @@ class DataTuple(TestTuple):
     def __init__(
         self, x: pd.DataFrame, s: pd.DataFrame, y: pd.DataFrame, name: Optional[str] = None
     ):
+        """Make a DataTuple."""
         super().__init__(x=x, s=s, name=name)
         self.__y: pd.DataFrame = y
 
@@ -209,6 +212,7 @@ class Prediction:
     """Prediction of an algorithm."""
 
     def __init__(self, hard: pd.Series, info: Optional[Dict[str, float]] = None):
+        """Make a prediction obj."""
         assert isinstance(hard, pd.Series), "please use pd.Series"
         self._hard = hard
         self._info = info if info is not None else {}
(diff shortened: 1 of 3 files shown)


tests/saving_data_test.py
```diff
@@ -1,20 +1,16 @@
-"""
-Test the saving data capability
-"""
+"""Test the saving data capability."""
```
olliethomas, line 1: There's a few of these. I have a pydocstyle pre-commit hook, so have to add this kind of thing.

tests/saving_data_test.py
```diff
@@ -48,10 +44,32 @@ def _script_command(self, train_path, _, pred_path):
     pd.testing.assert_series_equal(data_tuple.x["a1"], data_x.hard, check_names=False)
 
 
+def test_preidtions_loaded(temp_dir) -> None:
```
olliethomas, line 47:
```suggestion
def test_predictions_loaded(temp_dir) -> None:
```

pyproject.toml
```diff
@@ -43,27 +43,6 @@ skip-string-normalization = true
...
-  "setuptools",
-  "sklearn",
-  "tap",
-  "teext",
-  "torch",
-  "torchvision",
-  "tqdm",
-  "typing_extensions",
-]
+known_third_party = ["PIL", "black", "fairlearn", "git", "matplotlib", "mypy", "numpy", "pandas", "pylint", "pytest", "scipy", "seaborn", "setuptools", "sklearn", "tap", "teext", "torch", "torchvision", "tqdm", "typing_extensions"]
```
olliethomas, line 46: my isort plugin only started doing this in the pyproject.toml. It didn't do it in isort.config.
Maybe there's a setting I have that's not right

thomkeh:
Hm, but I thought I configured pydocstyle to not require docstrings in the `__init__` (requiring it only in the class itself)... maybe I didn't configure it in the right place.

ethicml/utility/data_structures.py
```diff
@@ -230,21 +234,33 @@ def info(self) -> Dict[str, float]:
...
-            return Prediction(hard=pd.Series(np.squeeze(data["hard"])))
+                return SoftPrediction(soft=pd.Series(np.squeeze(data["soft"])), info=info)
+            return Prediction(hard=pd.Series(np.squeeze(data["hard"])), info=info)
 
     def to_npz(self, npz_path: Path) -> None:
         """Save prediction as npz file."""
+        if self.info:
+            for v in self.info.values():
+                assert isinstance(v, float), "Info must be Dict[str, float]"
+            json_path = npz_path.parent / 'info.txt'
```
thomkeh, line 252: `.json` maybe? not that important though

```suggestion
            json_path = npz_path.parent / 'info.json'
```

olliethomas:
> Hm, but I thought I configured pydocstyle to not require docstrings in the `__init__` (requiring it only in the class itself)... maybe I didn't configure it in the right place.

or.... I have not made my pre-commit hook have the same requirements

thomkeh (approved):
I think it would have been possible to store the dictionary in the npz file together with the predictions, but it's fine.

olliethomas:
> I think it would have been possible to store the dictionary in the npz file together with the predictions, but it's fine.

this would be nicer - adding a .json file to the parent feels pretty bad

PR #404, opened 2020-12-16
Lipton

olliethomas:


ethicml/data/util.py
```diff
@@ -13,6 +16,27 @@
 ]
 
 
+def deprecated(func: Callable) -> Callable:
```
thomkeh, line 19: Please use this, otherwise all type information about the decorated function is lost:

```suggestion
_F = TypeVar("_F", bound=Callable[..., Any])


def deprecated(func: _F) -> _F:
```

pyproject.toml
```diff
@@ -43,6 +43,6 @@ skip-string-normalization = true
 
 [tool.isort]
 include_trailing_comma = "True"
-known_third_party = ["PIL", "black", "fairlearn", "git", "matplotlib", "mypy", "numpy", "pandas", "pylint", "pytest", "scipy", "seaborn", "setuptools", "sklearn", "tap", "teext", "torch", "torchvision", "tqdm", "typing_extensions"]
+known_third_party = ["PIL", "black", "git", "matplotlib", "mypy", "numpy", "pandas", "pylint", "pytest", "scipy", "seaborn", "setuptools", "sklearn", "tap", "teext", "torch", "torchvision", "tqdm", "typing_extensions"]
```
thomkeh, line 46: hm, why did you remove fairlearn?

olliethomas: errrr... .I didn't, autoflake does this - It must be because fairlearn isn't is a file-top import

olliethomas: I take it back, it's https://github.com/asottile/seed-isort-config and apparently it's out of date anyway. I'll remove it from the pre-commit

ethicml/data/util.py
```diff
@@ -12,6 +15,29 @@
...
+
+
+def deprecated(func: _F) -> _F:
+    """This is a decorator which can be used to mark functions as deprecated.
+
+    It will result in a warning being emitted when the function is used.
+    """
+
+    @functools.wraps(func)
+    def new_func(*args: Any, **kwargs: Any) -> Callable:
```
thomkeh, line 28: This return type is wrong actually. This is not returning a Callable but whatever `func` is returning.

Sorry that I spotted it so late!

olliethomas: Looking at the mypy docs, this needs tweaking a little so that the decorator does indeed return a callable https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators

PR #405, opened 2020-12-16
Don't check __init__ files for docstring style

thomkeh:


olliethomas:
Is this still draft?

thomkeh:
It could be merged; it's just not very comprehensive. The change is only applied to 3 files or so.

The overall idea was to move the section headings from the rst files to the `__init__.py` files themselves. In order to keep more of the documentation close to the code.

PR #407, opened 2020-12-17
a couple of lilnter things

olliethomas:


ethicml/algorithms/inprocess/zafar.py
```diff
@@ -18,6 +18,8 @@
 SUB_DIR_IMPACT = Path(".") / "disparate_impact" / "run-classifier"
 SUB_DIR_MISTREAT = Path(".") / "disparate_mistreatment" / "run_classifier"
 
+MAIN = "main.py"
```
thomkeh, line 21: This could even be given the `Final` type annotation (has to be imported from typing_extensions).

thomkeh (approved):

PR #409, opened 2021-01-06
replace dataset dataclass with regular python class

olliethomas:


thomkeh:
I think this can also be closed. As I understood it, the goal was to make `cont_features` private, which I did in #411.

olliethomas:
Sounds good!

PR #410, opened 2021-01-06
Prefer the capitalized dataset functions

thomkeh:
It seems like Oliver prefers them.


olliethomas:
I have no opinions on this

thomkeh:
I see! In case your wondering, I had come to this conclusion because when you added datasets, you've always added a capitalized version (like recently Lipton: https://github.com/predictive-analytics-lab/EthicML/blob/master/ethicml/data/tabular_data/lipton.py#L10 ). And the explanation that I came up with was that you liked the capitalized functions.

PR #412, opened 2021-01-25
Bump version on master in preparation for 0.2 release

thomkeh:


setup.py
```diff
@@ -4,8 +4,8 @@
 
 # We follow Semantic Versioning (https://semver.org/)
 _MAJOR_VERSION = "0"
-_MINOR_VERSION = "1"
-_PATCH_VERSION = "1"
+_MINOR_VERSION = "3"
```
olliethomas, line 7: ?

thomkeh: That's so that master is always ahead of any release.

thomkeh: If master were `0.2.0.dev`, then it would be considered older than `0.2.0`.

olliethomas: :D

PR #417, opened 2021-02-18
x referenced before assignment

olliethomas:


olliethomas:
btw, this (prior to the pr) is the same in fdm. How come it's not a problem?

thomkeh (approved):

thomkeh:
I think we never used this function in fdm.

PR #429, opened 2021-03-31
Try to replace setup.py with Pyproject

olliethomas:
Not sure that this will work first time


olliethomas:
@thomkeh do you have a preference about this one way or the other? At the moment when I use ethicml directly from github as a dependency in poetry, for some reason it doesn't get all the dependencies that ethicml needs, which is a faff. I'm hoping that by ethicml being managed by poetry too will make this problem go away

thomkeh:
@olliethomas sorry that I didn't respond! I prefer poetry!

thomkeh:
This will have conflicts with #433 :/

olliethomas:
let's get #433 in, we can deal with conflicts then

thomkeh:
@olliethomas rebased and conflicts solved

PR #430, opened 2021-04-27
add the UFRGS admissions dataset

olliethomas:


ethicml/data/tabular_data/admissions.py
```diff
@@ -0,0 +1,116 @@
...
+
+from ..dataset import Dataset
+from ..util import deprecated
+
+__all__ = ["Admissions", "admissions"]
+
+AdmissionsSplits = Literal["Gender"]
+
+
+@deprecated
```
thomkeh, line 51: @olliethomas why do you keep adding deprecated code? 😄 deprecated code is there to be removed! not added!

thomkeh: or are you trying to send the hidden message that you prefer the capitalized version of the function?

olliethomas: only because I don't know what's happening with the uppercase vs lowercase.

olliethomas: > or are you trying to send the hidden message that you prefer the capitalized version of the function?
I think it's safe to assume that there is no hidden message and that I'm mostly just copying the adult dataset file ever time I add a new dataset

PR #450, opened 2021-07-05
Add option to load labels into the input feature dataframe

thomkeh:
Also, 3 tests were marked as slow which aren't actually slow.


ethicml/data/dataset.py
```diff
@@ -152,15 +161,17 @@ def load(self, ordered: bool = False) -> DataTuple:
         assert isinstance(dataframe, pd.DataFrame)
 
         feature_split = self.feature_split if not ordered else self.ordered_features
-        feature_split_x = feature_split["x"]
+        if keep_labels_as_features:
+            feature_split_x = feature_split["x"] + feature_split["s"] + feature_split["y"]
+        else:
+            feature_split_x = feature_split["x"]
```
thomkeh, line 167: (the above is the main change)

thomkeh:
I can think of some edge cases where this fails as implemented right now, but I think it's good enough for now.

ethicml/data/dataset.py
olliethomas: should there be the option to just have S, but not Y, included in the features?

PR #455, opened 2021-08-14
make the dataset splits enums

olliethomas:


ethicml/data/tabular_data/admissions.py
```diff
@@ -37,19 +37,31 @@
...
 __all__ = ["admissions"]
 
-AdmissionsSplits = Literal["Gender"]
 
+class AdmissionsSplits(Enum):
+    GENDER = "Gender"
+    CUSTOM = "Custom"
+
+
+VALID_STRS = Literal[tuple([e.value for e in AdmissionsSplits])]  # type: ignore[misc]
```
thomkeh, line 55: You may as well use `str`. No static type checker can parse this (I think?).

ethicml/data/tabular_data/admissions.py
```diff
@@ -37,19 +37,31 @@
...
 
 from ..dataset import Dataset
 from ..util import flatten_dict
 
 __all__ = ["admissions"]
 
-AdmissionsSplits = Literal["Gender"]
 
+class AdmissionsSplits(Enum):
+    GENDER = "Gender"
```
thomkeh, line 51: All caps for the values?

thomkeh (approved):

olliethomas:
The motivation is that when I call for example the `adult` function and pass `Race`, I get "expected Literal not str".

ethicml/data/tabular_data/admissions.py
olliethomas: Apparently all caps for the name is the done thing according to pylint. I'm not sure I like it

ethicml/data/tabular_data/admissions.py
olliethomas: Yeah...

PR #457, opened 2021-09-13
Simplify/generalise compute_instance_weights.

MylesBartlett:


thomkeh (approved):
Seeing as this passes the tests, I conclude that the code must be correct

PR #458, opened 2021-09-27
make is so that there is a flag to make the sensitive attribute name is consistent across datasets.

olliethomas:


thomkeh (approved):
`run_metrics()` in `evaluators/evaluation.py` should also get a new argument to pass it on to `metric_per_sensitive_attribute`

PR #459, opened 2021-09-27
poetry update; add parsable to model inits; prefer Enums

olliethomas:


olliethomas:
this won't create a new release on it's own

thomkeh:
I think this has been superseded by #590 .

PR #460, opened 2021-09-27
try to emulate the poetry setup from conduit

olliethomas:
not entirely sure I've done this right tbh

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -131,7 +131,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: .venv
-          key: venv2-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous_integration.yml
```diff
@@ -126,3 +95,61 @@ jobs:
...
+
+      #----------------------------------------------
+      #       load cached venv if cache exists
+      #----------------------------------------------
+      - name: Set up Poetry cache for Python dependencies
+        id: cached-poetry-dependencies
+        uses: actions/cache@v2
+        with:
+          path: .venv
+          key: venv2-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
```
thomkeh, line 134: I only put the 2 here to invalidate a specific cache in palkit. we should probably remove it
```suggestion
          key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
```

PR #469, opened 2021-12-10
Add Income dataset from Retiring Adult

olliethomas:


ethicml/__init__.py
```diff
@@ -1,4 +1,6 @@
 """EthicML."""
+from ethicml.metrics.per_sensitive_attribute import *
```
thomkeh, line 2: This import looks wrong to me, but if it works...

ethicml/metrics/average_odds.py
```diff
@@ -1,11 +1,17 @@
 """For assessing Average Odds Difference metric."""
 
 
-from kit import implements
+from ranzen import implements
```
thomkeh, line 5: I would also do a relative import here

ethicml/metrics/cv.py
```diff
@@ -1,7 +1,11 @@
 """For assessing Calder-Verwer metric, :math:`1-(P(Y=1|S=1)-P(Y=1|S!=1))`."""
 
-from kit import implements
+from ranzen import implements
 
+from ethicml.metrics.per_sensitive_attribute import (
```
thomkeh, line 5: And here

thomkeh (approved):
Moving per_sensitive_attribute was a really good idea. Why didn't we do this before?

PR #472, opened 2021-12-22
add separate fit and test functions

olliethomas:
Added separate fit and predict methods for the InProcessing models.

This is mostly just splitting up what's already packaged together in various `train_and_predict` functions.

The bit that needs double checking is for the async algos. Here a model save location now gets passed, then the trained model is just dumped there. I'm not sure if there's a better way to do this or not. I imagine that there is.

There's now quite a lot of duplication. This should really be sorted....

### DIFF
--- a/ethicml/algorithms/inprocess/in_algorithm.py
+++ b/ethicml/algorithms/inprocess/in_algorithm.py
@@ -1,4 +1,6 @@
 """Abstract Base Class of all algorithms in the framework."""
+from __future__ import annotations
+
 from abc import abstractmethod
 from pathlib import Path
 from tempfile import TemporaryDirectory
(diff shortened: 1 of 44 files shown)


olliethomas:
this is useful if you want to train a model and then evaluate on a bunch of different test sets

abarrainkua (approved):

abarrainkua (approved):

thomkeh:
As we discussed before, the whole async stuff should be ripped out anyway and replaced by something based on `ray`.

Apart from that, this seems fine.

ethicml/algorithms/inprocess/in_algorithm.py
```diff
@@ -15,10 +17,32 @@
...
     """Abstract Base Class for algorithms that run in the middle of the pipeline."""
 
-    def __init__(self, name: str, is_fairness_algo: bool = True):
-        super().__init__(name=name)
+    def __init__(self, name: str, seed: int, is_fairness_algo: bool = True):
+        super().__init__(name=name, seed=seed)
         self.__is_fairness_algo = is_fairness_algo
 
+    @abstractmethod
+    def fit(self, train: DataTuple) -> InAlgorithm:
```
thomkeh, line 25: This would be the perfect opportunity to use `-> Self` but it's not widely supported yet: https://github.com/python/typeshed/issues/6300

thomkeh (approved):

PR #474, opened 2021-12-30
Try to turn datasets into dataclasses

thomkeh:
This would obviate the need for (neo)configen when using EthicML with hydra.

I'm not sure whether this is a good idea. I only did it for Adult and COMPAS and I'm sure lots of tests will fail.


olliethomas:
I'm all for this. Should this be merged, or wait until all the datasets are done?

thomkeh:
Great! Let's merge now then. I can do the rest later.

PR #476, opened 2022-01-04
aif converter

olliethomas:
- add method to load data in the AIF360 format
- mypy

This adds a small function that loads data in the AIF360 format so that you can use with their algorithms


thomkeh (approved):
A test for _from_dummies would be nice but it's fine.

PR #478, opened 2022-01-05
Make the new dataset dataclasses hydra-compatible

thomkeh:
This meant removing the `Union`. However, we can keep the `Union` on the functional versions. This should allow us to remain backwards-compatible.

I also added a test with `OmegaConf` that ensures that the dataclasses can indeed be used by hydra directly.


thomkeh:
`omegaconf` is only a dev and CI dependency.

olliethomas (approved):

PR #479, opened 2022-01-05
Replace custom async code with ray

thomkeh:
I hope this allows us to remove a lot of code. Currently it's not removing that much, because many algorithms are written subprocess calling in mind and have to be restructured.

I promise I will get to fixing the InstalledModel afterwards :smile:


olliethomas:
Ray definitely feels better than subprocess. Is the idea that there is then no `run_async` anymore?

e.g.
```
import ethicml as em
train, test = em.train_test_split(em.Adult().load())
model = SVMAsync(dir='/tmp')
predictions = model.run(train, test)
print(sum(predictions.hard))
```

thomkeh:
Yes, no `run_async` anymore. `run_async` was just there so that `run_in_parallel` could run more than one algorithm at a time. But with ray, we have true parallelism now, so even algorithms that don't define `run_async` can run in parallel now. So, `run_async` is pointless.

To be honest, I have the impression no one is using `run_in_parallel` (or `evaluate_models_async`) anyway, so my change won't really affect anyone. (Whenever one of the async models was used, you were calling most likely `run()` which in turn does `run_blocking(self.run_async(...))`. This is now how it always is.)

PR #483, opened 2022-01-07
Turn all datasets into dataclasses

thomkeh:
Well, except for ACS Income, because that one seems complicated.


thomkeh:
I briefly tried to turn the ACS datasets into dataclasses as well, but it was annoying. Let me know if you need that @olliethomas . I will merge this now anyway, because nothing here should be controversial – just a continuation of #474 .

PR #489, opened 2022-01-12
Define labels for confusion-matrix metrics

olliethomas:


olliethomas:
This PR adds an option to metrics that are based on a confusion matrix to specify the labels that exist in a dataset, rather than always inferring these. Now they are inferred by default, but can be overridden by passing a list of int.

thomkeh (approved):
LGTM

PR #492, opened 2022-01-13
Tried to make it so that aif and em are interoperable. 

olliethomas:
Turns out that it's totally possible to switch between EM and AIF datasets, providing that you do a lot of the definitions at runtime.
The problem is that there are some fundamental design differences, and I don't know how to get past them, so putting this here so that everyone can see what I've done.

### DIFF
--- a/ethicml/algorithms/preprocess/pre_algorithm.py
+++ b/ethicml/algorithms/preprocess/pre_algorithm.py
@@ -32,15 +32,14 @@ class PreAlgorithm(Algorithm):
         self._out_size = out_size
 
     @abstractmethod
-    def fit(self, train: DataTuple) -> Tuple[PreAlgorithm, DataTuple]:
+    def fit(self, train: DataTuple) -> PreAlgorithm:
         """Generate fair features with the given data.
 
         Args:
             train: training data
-            test: test data
 
         Returns:
-            a tuple of the pre-processed training data and the test data
+            a trained model
         """
 
     @abstractmethod
@@ -48,11 +47,10 @@ class PreAlgorithm(Algorithm):
         """Generate fair features with the given data.
 
         Args:
-            train: training data
             test: test data
 
         Returns:
-            a tuple of the pre-processed training data and the test data
+            transformed test data
         """
 
     @abstractmethod
--- a/ethicml/__init__.py
+++ b/ethicml/__init__.py
@@ -23,6 +23,7 @@ from .algorithms.preprocess.pre_algorithm import *
 from .algorithms.preprocess.upsampler import *
 from .algorithms.preprocess.vfae import *
 from .algorithms.preprocess.zemel import *
+from .algorithms.preprocess.zemel_aif import *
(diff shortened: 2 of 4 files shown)


olliethomas:
Is there a way to turn checks off? I know that this will fail.

@thomkeh 
couple things:
- the shape of the returned dataset is len(x) + len(s).
- I can drop the S feature and the shape will be right, but the bigger problem is that this shouldn't really be available (IMO) during training of the algorithm in the first place.

thomkeh:
Okay, so the problem is that AIF360 expects s as part of the input features, but if you add s to x, then it is hard after the transformation to separate s and x again?

olliethomas:
> Okay, so the problem is that AIF360 expects s as part of the input features, 
It does

> but if you add s to x, then it is hard after the transformation to separate s and x again?
No, it's possible. It's a bit awkward because you have a numpy array and a feature list, but that's not so much of a problem (assuming that the order of the list and the features align)

The problem is that the learning procedures in aif learns a mapping `f: (x,s) -> y`
In contrast we learn `f: x -> y`. While sometimes `s` is used during training e.g. in a loss function, we don't treat `s` as just another column in `x`
If the purpose of this is to be able to call AIF360 models from EM, then suddenly there is not a fair comparison.

thomkeh:
Hm, this is an interesting question. We do make `s` available to the algorithms, I think (`TestTuple` has both `x` and `s`). But they usually don't make use of it. Why did we do it like that? :thinking:

olliethomas:
Why do we do it like that as in, why do we _not_ use `s` as a feature?

Because outcomes shouldn't be based on sensitive attributes as to avoid claims of [disparate treatment](https://en.wikipedia.org/wiki/Disparate_treatment). Or this was the thinking anyway.

Any model that is trained as `(x,s) -> y` will also need `(x,s)` during inference.

ethicml/algorithms/preprocess/pre_algorithm.py
```diff
@@ -32,27 +32,25 @@ def __init__(self, name: str, seed: int, out_size: Optional[int]):
         self._out_size = out_size
 
     @abstractmethod
-    def fit(self, train: DataTuple) -> Tuple[PreAlgorithm, DataTuple]:
```
thomkeh, line 35: why was the return type like that?

thomkeh (commented):
in general I have no objections to merging this. but if you think a better design is possible, we can hold off.

ethicml/algorithms/preprocess/pre_algorithm.py
olliethomas: 🤔  it should only really be the `PreAlgorithm`. I think I did it with both because the training set was always transformed so it feels like a waste to not return it... but this does make the api ugly...

thomkeh: is the solution to introduce a `fit_and_transform()`?

PR #493, opened 2022-01-21
Replace count_true with np.count_nonzero

### LINKED ISSUE #452: Replace `count_true` in the tests with `np.count_nonzero`

thomkeh:
closes #452


olliethomas (approved):

thomkeh:
An even better solution would be to use [np.unique with return_counts=True](https://numpy.org/doc/stable/reference/generated/numpy.unique.html#numpy.unique) but whatever.

PR #518, opened 2022-03-11
Turn metrics into dataclasses

thomkeh:
Same as before with the datasets and algorithms.

### DIFF
--- a/ethicml/metrics/metric.py
+++ b/ethicml/metrics/metric.py
@@ -15,8 +15,6 @@ class Metric(Protocol):
 
     apply_per_sensitive: bool
     """Whether the metric can be applied per sensitive attribute."""
-    # name: str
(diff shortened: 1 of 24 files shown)


ethicml/metrics/metric.py
```diff
@@ -1,21 +1,22 @@
...
-    _name: str = "<unnamed metric>"
+class Metric(Protocol):
+    """Base class for all metrics."""
 
-    def __init__(self, pos_class: int = 1) -> None:
-        self.positive_class = pos_class
+    apply_per_sensitive: bool
+    """Whether the metric can be applied per sensitive attribute."""
+    # name: str
+    # """Name of the metric."""
```
thomkeh, line 19:
```suggestion
```

PR #529, opened 2022-03-14
Auto merge dependabot

olliethomas:


olliethomas:
I have no idea why the dummy suite isn't being triggered here

.github/workflows/dependabot_auto.yml
```diff
@@ -1,5 +1,9 @@
 name: Dependabot auto-approve
-on: pull_request
+on:
+  workflow_run:
+    workflows: [continuous_integration]
+    types:
+      - completed
```
thomkeh, line 6: Does "completed" imply *successful*?

PR #530, opened 2022-03-14
add guard that ci workflow passes successfully

olliethomas:
I'm tempted to make the workflow dir trigger the dummy CI rather than the real one


thomkeh:
You can certainly exclude this specific yaml file from the normal CI.

thomkeh (approved):

PR #532, opened 2022-03-15
add some more text to the cross validator docstrings

olliethomas:


thomkeh:
some code examples (in the module docstring, for example) would be great, but I know it's a lot of work

ethicml/evaluators/cross_validator.py
```diff
@@ -173,7 +194,14 @@ def run_async(self, train: DataTuple, measures: Optional[List[Metric]] = None) -
         return CVResults(compute_scores_and_append.results, self.model)
 
     def run(self, train: DataTuple, measures: Optional[List[Metric]] = None) -> CVResults:
-        """Run the cross validation experiments."""
+        """Run the cross validation experiments asynchronously.
```
thomkeh, line 197:
```suggestion
        """Run the cross validation experiments.
```

thomkeh (approved):

olliethomas:
btw, I don't really know if this is how do you do code examples in ReST. I looked it up and it seems like this is right, but 🤷‍♂️

PR #535, opened 2022-03-16
Create 404.md

olliethomas:
I have no idea if this will work.

Just trying to follow https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-custom-404-page-for-your-github-pages-site

### DIFF
--- a/docs/source/404.md
+++ b/docs/source/404.md
@@ -1,0 +1,8 @@
+---
+permalink: /404.html
+---
+
+Oh no! This page doesn't exist, or may have moved. 
+You can get back to the documentation for EthicML [here](https://wearepal.ai/EthicML/).
+
+Or, navigate yourself to https://wearepal.ai/EthicML/


thomkeh:
No, that will not work. It should be `.rst`. And no need to specify the permalink, I think.

thomkeh:
(why should it be `.rst`? Because this will be processed by sphinx and not by GitHub's Jekyll)

olliethomas:
edited to be .rst and permalink removed

olliethomas:
I'm really not sure why this hasn't triggered the dummy CI

olliethomas:
> Well, nothing seems to be triggered. Maybe I made a mistake in the workflow config.

I think GH is just running slowly, based on slack taking several minutes to reflect PRs being opened. 
The 404 isn't showing, but perhaps this is also due to GH being slow to reflect changes rather than not working

PR #537, opened 2022-03-16
remove Postprocess from evaluate_models as its not used. Store hyperparams in results table. Append Hyperparams to K&C name.

olliethomas:


ethicml/algorithms/inprocess/logistic_regression.py
```diff
@@ -30,6 +30,9 @@ class LR(InAlgorithm):
     seed: int = 888
     is_fairness_algo = False
 
+    def __post_init__(self):
+        self._hyperparameters = {"C": self.C}
```
thomkeh, line 34: Would have preferred to do this with a property instead of with `__post_init__` but I don't see an elegant way to do it here.

thomkeh (approved):
Hm, I guess this is fine. I personally would have preferred to go all in on dataclasses so that you can just call `asdict()` on the algorithm object to get the hyperparameters as a dict (`seed` should not be a parameter of the algorithm instantiation at all – it should be passed in through `run()` explicitly), but this also fine.

PR #539, opened 2022-03-17
add a default arg for dir. Also, make most model inits be kwarg-only

olliethomas:


ethicml/algorithms/inprocess/agarwal_reductions.py
```diff
@@ -52,6 +53,7 @@ def __init__(
             raise ValueError(f"results: classifier must be one of {VALID_MODELS!r}.")
         self.seed = seed
         self.is_fairness_algo = True
+        dir = dir if dir is not None else "."
```
MylesBartlett, line 56:
```suggestion
        dir = ". if dir is None else dir
```

ethicml/algorithms/preprocess/beutel.py
```diff
@@ -17,9 +17,9 @@ class Beutel(PreAlgorithmAsync):
 
     def __init__(
         self,
-        dir: Union[str, Path],
         fairness: FairnessType = "DP",
         *,
+        dir: Optional[Union[str, Path]] = None,
```
thomkeh, line 22: breaking change :scream:

thomkeh (approved):
I think I would have done
```python
dir: Union[str, Path] = "."
```
but maybe you have a good reason not to do it that way.

PR #540, opened 2022-03-17
Scaler should be used in evaluate models async

olliethomas:


olliethomas:
no - I just added it based on the matching line in the non-async version. I then just assumed that the tests would fail if this was bad.

thomkeh:
Ok, let's assume it works then.

PR #542, opened 2022-03-17
add a property of Kamiran that returns the instance weights of the gr…

olliethomas:
…oups


thomkeh (approved):
Maybe add
```python
self.group_weights: Optional[Dict[int, Dict[str, float]]] = None
```
to the `__init__()`.

PR #547, opened 2022-03-17
Remove vision

olliethomas:


olliethomas:
the `merge-main` branch should be merged in first. Then the diff would be clearer

MylesBartlett:
The end of an era  🌆

olliethomas:
> The end of an era 🌆

It will still be a while until it's merged into main (but hopefully not that long)

PR #559, opened 2022-03-25
Remove TAP and reduce code duplication

thomkeh:


thomkeh:
I will run the slow tests as well.

PR #561, opened 2022-03-26
Add RobustAccuracy Metric

olliethomas:
I'm under no impression that this is good. Indexing by S in EM should be so much more elegant

### DIFF
--- a/ethicml/metrics/accuracy.py
+++ b/ethicml/metrics/accuracy.py
@@ -55,11 +55,11 @@ class RobustAccuracy(SklearnMetric):
     def score(self, prediction: Prediction, actual: DataTuple) -> float:
         scores = []
         for _s in actual.s[actual.s.columns[0]].unique():
-            preds_subset = Prediction(hard=prediction.hard[actual.s[actual.s.columns[0]] == _s])
+            preds_subset = Prediction(hard=prediction.hard[actual.s.iloc[:, 0] == _s])
             data_subset = DataTuple(
-                x=actual.x[actual.s[actual.s.columns[0]] == _s],
-                s=actual.s[actual.s[actual.s.columns[0]] == _s],
-                y=actual.y[actual.s[actual.s.columns[0]] == _s],
+                x=actual.x[actual.s.iloc[:, 0] == _s],
+                s=actual.s[actual.s.iloc[:, 0] == _s],
+                y=actual.y[actual.s.iloc[:, 0] == _s],
             )
             scores.append(super().score(preds_subset, data_subset))
         return min(scores)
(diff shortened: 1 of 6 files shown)


MylesBartlett (approved):

thomkeh:
Instead of
```
data.s[data.s.colums[0]]
```
you can do
```
data.s.iloc[:, 0]
```

ethicml/metrics/accuracy.py
```diff
@@ -41,3 +41,25 @@ class F1(SklearnMetric):
...
+    @implements(SklearnMetric)
+    def score(self, prediction: Prediction, actual: DataTuple) -> float:
+        scores = []
+        for _s in actual.s[actual.s.columns[0]].unique():
+            preds_subset = Prediction(hard=prediction.hard[actual.s[actual.s.columns[0]] == _s])
+            data_subset = DataTuple(
+                x=actual.x[actual.s[actual.s.columns[0]] == _s],
+                s=actual.s[actual.s[actual.s.columns[0]] == _s],
+                y=actual.y[actual.s[actual.s.columns[0]] == _s],
+            )
```
thomkeh, line 63: Hm, I feel we must have a method for applying a mask to a data tuple

olliethomas:
done - qq, do you know why the sklearnmetric is a tuple?

ethicml/metrics/accuracy.py
```diff
@@ -41,3 +41,25 @@ class F1(SklearnMetric):
     sklearn_metric = (f1_score,)
     apply_per_sensitive: ClassVar[bool] = True
     _name: ClassVar[str] = "F1"
+
+
+@dataclass
+class RobustAccuracy(SklearnMetric):
+    """Classification accuracy."""
```
thomkeh, line 48:
```suggestion
    """Robust accuracy."""
```

ethicml/metrics/accuracy.py
olliethomas: yeah, I'd have though so, we do it in enough places

thomkeh:
Yes, it's explained here: https://github.com/predictive-analytics-lab/EthicML/blob/main/ethicml/metrics/accuracy.py#L20

thomkeh:
But I'm not sure you gained much by subclassing SklearnMetric

olliethomas:
So, technically this is a breaking change now as I've renamed `DataTuple.get_subset`, but i don't think anyone uses this directly anyway

thomkeh:
So, the thing I was thinking of was `apply_to_joined` but that's probably overkill here. It would look like this:
```
subset = datatup.apply_to_joined(lambda df: df.loc[datatup.s == _s])
```

PR #587, opened 2022-04-13
Add aliases for dataset splits within the dataset classes

thomkeh:
See here what it looks like:

https://github.com/predictive-analytics-lab/EthicML/blob/enum-aliases/tests/hydra_compatibility_test.py#L13


MylesBartlett (approved):

thomkeh:
Now I'm not sure whether to rather go the easier route while accepting that the alias won't work for type annotations.

PR #589, opened 2022-04-13
Turn SVM model class into a dataclass

thomkeh:
This technically breaks backwards compatibility with respect to the algorithm name (which is why I didn’t do it before), but I don’t really think it’s that bad.

This will not help with your configen problem though, Oliver. For that, we'd need to convert the Literal to an enum.

### DIFF
--- a/tests/models_test/inprocess_test/models_inprocessing_test.py
+++ b/tests/models_test/inprocess_test/models_inprocessing_test.py
@@ -78,7 +78,7 @@ INPROCESS_TESTS = [
     InprocessTest(name="Majority", model=Majority(), num_pos=80),
     InprocessTest(name="MLP", model=MLP(), num_pos=43),
     InprocessTest(name="Oracle", model=Oracle(), num_pos=41),
-    InprocessTest(name="SVM", model=SVM(), num_pos=45),
+    InprocessTest(name="SVM (rbf)", model=SVM(), num_pos=45),
     InprocessTest(name="SVM (linear)", model=SVM(kernel="linear"), num_pos=41),
 ]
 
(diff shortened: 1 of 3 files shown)


thomkeh:
I don't know why slack wasn't triggered.

olliethomas:
> I don't know why slack wasn't triggered.

was the pr opened before the label was added?

thomkeh:
> > I don't know why slack wasn't triggered.
> 
> was the pr opened before the label was added?

I guess it must have been like that, but I did use the github CLI with `gh pr -l code-change` so I thought that would add it immediately.

tests/models_test/inprocess_test/models_inprocessing_test.py
```diff
@@ -78,7 +78,7 @@ class InprocessTest(NamedTuple):
     InprocessTest(name="Majority", model=Majority(), num_pos=80),
     InprocessTest(name="MLP", model=MLP(), num_pos=43),
     InprocessTest(name="Oracle", model=Oracle(), num_pos=41),
-    InprocessTest(name="SVM", model=SVM(), num_pos=45),
+    InprocessTest(name="SVM (rbf)", model=SVM(), num_pos=45),
```
thomkeh, line 81: see here

tests/models_test/preprocess_test/models_preprocessing_test.py
```diff
@@ -165,7 +165,7 @@ def test_threaded_pre(toy_train_test: TrainTestPair, model: PreAlgorithm, name:
 
     svm_model: InAlgorithm = SVM()
     assert svm_model is not None
-    assert svm_model.name == "SVM"
+    assert svm_model.name == "SVM (rbf)"
```
thomkeh, line 168: and here

thomkeh (commented):
So, are you fine with this breakage of backwards compatibility in the name? Alternatively, we could also check whether the kernel is "rbf" and then exclude it from the name.

olliethomas:
Maybe this should be merged into `next`.... 🤔

PR #596, opened 2022-04-19
Fix evaluate models

thomkeh:


ethicml/evaluators/evaluate_models.py
```diff
@@ -304,6 +308,6 @@ def _gather_metrics(
         # put old results before new results -> prepend=True
         aggregator.append_from_csv(csv_file, prepend=True)
         aggregator.save_as_csv(csv_file)
-        all_results.append_df(aggregator.results)
+        all_results.append_df(results_df)
```
thomkeh, line 311: this was the main trick

tests/visualisation_test.py
```diff
@@ -96,8 +95,7 @@ def test_plot_evals():
     # plot with metrics
     figs_and_plots = plot_results(results, Accuracy(), ProbPos())
     # num(datasets) * num(preprocess) * num(accuracy combinations) * num(prop_pos combinations)
-    assert len(figs_and_plots) == 2 * 2 * 1 * 2 + 4  # TODO: this +4 should be FIXED,
-    # it matches the column name containing a hyphen as a DIFF metric.
```
thomkeh, line 100: I really don't understand why this is suddenly fixed...

thomkeh:
I have realized that this is actually still wrong but it will be fine as long as there is no previous data. #597 is closer to being correct but also still not quite correct. However, I don't really feel like this is important enough to spend more time on.

PR #598, opened 2022-04-20
Replace ray with joblib.Parallel

### LINKED ISSUE #595: Replace ray with joblib.Parallel
`ray` is complete overkill for just running things in parallel. [joblib.Parallel](https://joblib.readthedocs.io/en/latest/parallel.html) will do just fine. (The only thing that joblib can't do is manage access to GPUs, but I don't think we need that.)

thomkeh:
closes #595

I hope this will speed up the CI.


olliethomas (approved):

olliethomas:
Closes #595 ?

PR #590, opened 2022-04-14
remove some literals

olliethomas:
Should cover all of the model inits... I think

PR #603, opened 2022-04-29
preds hard check

olliethomas:
- Add next version branch. Update workflows to reflect this new branch
- Remove vision (#547)
- Bump version on the `next` branch (#549)
- Update lock file after solving merge conflicts (#581)
- Turn SVM model class into a dataclass (#589)
- Replace most Literals with Enums (#590)
- Replace ray with joblib.Parallel
- Update dependencies
- Convert docstrings from google format to sphinx format
- Fix example
- Fix sphinx build
- Convert a few stragglers
- Write Default: instead of Default value =
- Use GitHub Copilot to generate docstrings
- A few more docstring improvements
- Remove AIF360 dependency
- Disable the AIF360 test
- preds.hard should be guaranteed to be an int


olliethomas:
loooool. This was meant for main. i need to learn how to use `gh` command line tools

PR #606, opened 2022-04-29
add run methods to models

olliethomas:
The problem is that joblib.Memory saves the results cased on the function name https://joblib.readthedocs.io/en/latest/memory.html#gotchas

And when the method is in the parent, then the function name is the same. So this tries to deal with that.


olliethomas:
> I can't say I love this...

I don't wither, but what's the alternative?

thomkeh:
Even if we do this, this PR has too many conflicts to be useful.

PR #615, opened 2022-05-03
add option to specify test set size

olliethomas:
tidied up the tests a little while I was at it.


olliethomas:
@thomkeh there's not much to this, but right now I would find it helpful to be able to specify the test set size. I don't love these args where only one can be set, I guess I could make 2 separate functions, but the name is of the current function is too good to mess with.

thomkeh (approved):
Sure, looks good

PR #624, opened 2022-05-06
Change default of `delete_prev` to `True`

thomkeh:


olliethomas:
this is a good idea. The motivation behind it being false was so that you don't accidentally delete important work. But most methods are pretty fast, so I don't that's as much of a problem as I thought it would be

PR #625, opened 2022-05-06
closes #623.

### LINKED ISSUE #623: Vary model seed in addition to datasplit seed in `evaluate_models()`
Now that #538 has been addressed, this is the next step.

olliethomas:


ethicml/evaluators/evaluate_models.py
```diff
@@ -206,6 +213,7 @@ def evaluate_models(
                 train = train.get_n_samples()
             train = train.replace(name=f"{train.name} ({split_id})")
             data_splits.append(TrainTestPair(train, test))
+            model_seeds.append(0 if repeat_on == "data" else split_id)
```
thomkeh, line 216: maybe we should store the model seeds in the `TrainTestPair`? that might be an easier way to get the information where it needs to go

thomkeh: ah no, I see that `TrainTestPair` is also used in the tests where it doesn't make sense to have a seed

ethicml/evaluators/parallelism.py
```diff
@@ -76,22 +81,27 @@ def arrange_in_parallel(
...
+def _run(algo: Algorithm[_RT], train_test_pair: TrainTestPair, seed: int) -> _RT:
     train, test = train_test_pair
     # do the work
     try:
         result: _RT = algo.run(train, test)
+        if isinstance(result, Prediction):
+            result.info["model_seed"] = seed
     except RuntimeError:
         result = Prediction(hard=pd.Series([np.NaN] * len(test)))
     return result
```
olliethomas, line 107: Just realised that this will not work in the case that a PreProcess method fails.

olliethomas: I'll add an issue as it's a pretty rare event

thomkeh (approved):
I definitely approve of the general idea. You just need to make the tests pass

ethicml/evaluators/parallelism.py
thomkeh: Right. Doesn't mypy complain?

olliethomas: no. I don't think that it can know about this.

PR #626, opened 2022-05-06
add graceful failure... well at least complete and don't crash

olliethomas:


thomkeh (approved):
That seems reasonable

PR #632, opened 2022-05-06
tight layout

olliethomas:


olliethomas:
I don't actually know if this will do what I want it to... but atm the returned figures are cropped when I log to w&b, and plt.tight_layout or fig.tight_layout don't seem to work at my end

PR #650, opened 2022-05-20
Update OmegaConf and remove PyTorch from dependencies

thomkeh:
Having PyTorch in the dependencies only causes problems. You won't get the right version anyway with poetry and just the presence of PyTorch in the dependencies slows down `poetry update` enormously. It's just not worth it.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -126,7 +126,9 @@ jobs:
       #----------------------------------------------
       - name: Install dependencies
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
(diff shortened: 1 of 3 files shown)


olliethomas:
I'm a bit surprised that all the `import torch` statements don't cause problems

thomkeh:
torch is still installed in the CI. I'm just doing it manually so that we get the CPU version:

https://github.com/predictive-analytics-lab/EthicML/blob/200aa8e4faec3bebbd70fb1ad6429812f92606ce/.github/workflows/continuous_integration.yml#L131

thomkeh:
Hm, though I should maybe pin the pytorch version.

PR #653, opened 2022-05-23
issue 647

olliethomas:
- wip
- make the soft predictions return a probability for each class


olliethomas:
The idea for this PR is to use SoftPrediction more. For many classifiers we have the option to get a soft output, but we don't. So that was the idea. This led a little but down a rabbit hole as we're using "is the first dim >= 0.5" to determine the class label, when clearly argmax is much better for this. This also led to the Non-Binary test data being regenerated.

ethicml/algorithms/preprocess/calders.py
```diff
@@ -2,15 +2,17 @@
...
 from ranzen import implements
 
 from ethicml.utility import DataTuple, SoftPrediction, TestTuple, concat_dt
 
-from ..inprocess.logistic_regression import LRProb
 from .pre_algorithm import PreAlgorithm, T
 
 __all__ = ["Calders"]
 
+from ..inprocess.logistic_regression import LR
```
thomkeh, line 14: Should be above `__all__`, no?

ethicml/algorithms/preprocess/upsampler.py
```diff
@@ -9,11 +9,12 @@
 
 from ethicml.utility import DataTuple, SoftPrediction, TestTuple
 
-from ..inprocess.logistic_regression import LRProb
 from .pre_algorithm import PreAlgorithm, T
 
 __all__ = ["Upsampler", "UpsampleStrategy"]
 
+from ..inprocess.logistic_regression import LR
```
thomkeh, line 16: Same

thomkeh (approved):
Makes sense that soft predictions can't be a pd.Series in the non-binary case.

PR #659, opened 2022-05-24
Use single DataFrame as underlying storage for DataTuple

### LINKED ISSUE #379: Store data tuple internally as a single pd.DataFrame
My dream is to store the datatuple internally as a single dataframe and the property functions like `data.s` do something like this:
```python
return self._internal_df[self.s_name]
```
i.e., the datatuple class returns columns from this dataframe on demand.

Storing everything as a single dataframe would save a lot of hassle. At the moment we have to constantly join and split the components.

Pseudo code:
```python
@dataclass
class DataTuple:
    all_data: pd.DataFrame
    x_columns: List[str]
    s_column: str
    y_column: str
		
    @property
    def x(self) -> pd.DataFrame:
        return self.all_data[self.x_columns]
		
    @property
    def s(self) -> pd.Series:
        return self.all_data[self.s_column]
	
    @property
    def y(self) -> pd.Series:
        return self.all_data[self.y_column]
```

thomkeh:
closes #379


ethicml/algorithms/preprocess/calders.py
```diff
@@ -75,7 +75,7 @@ def _calders_algorithm(
     data: Dict[Tuple[int, int], DataTuple] = {}
     for s, y in groups:
         s_y_mask = (dataset.s == s) & (dataset.y == y)
-        data[(s, y)] = DataTuple(
+        data[(s, y)] = DataTuple.from_x_s_and_y(
```
MylesBartlett, line 78:
```suggestion
        data[(s, y)] = DataTuple.from_x_s_and_y(
```
Maybe drop the 'and' for the sake of brevity?

thomkeh: done.

PR #663, opened 2022-05-26
Replace pandas with polars

thomkeh:
I just did search&replace. This does not work yet.


thomkeh:
This will require more work and there is little point in solving the merge conflicts here.

thomkeh:
This will require more work and there is little point in solving the merge conflicts here.

PR #664, opened 2022-05-26
Don't let DataTuple inherit from TestTuple

thomkeh:
And also introduce the new `LabelTuple`.

The names are all subject to change. I can easily rename them with my editor.

Myles might be excited about this:

```python
preds = Prediction(hard=pd.Series([0, 0, 1, 1]))
targets = LabelTuple.from_np(
    s=np.array([0, 1, 1, 1]), y=np.array([0, 1, 0, 1]), s_name="s", y_name="y"
)
score = metric.score(preds, targets)
```

### DIFF
--- a/ethicml/algorithms/inprocess/fair_dummies.py
+++ b/ethicml/algorithms/inprocess/fair_dummies.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from typing import List
-from typing_extensions import Final, Literal, TypedDict
+from typing_extensions import Literal, TypedDict
 
 from ranzen import implements
 
--- a/ethicml/algorithms/inprocess/fair_dummies.py
+++ b/ethicml/algorithms/inprocess/fair_dummies.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from typing import List
-from typing_extensions import Literal, TypedDict
+from typing_extensions import Final, Literal, TypedDict
 
 from ranzen import implements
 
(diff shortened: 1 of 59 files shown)


ethicml/utility/data_structures.py
```diff
@@ -256,6 +287,150 @@ def from_npz(cls, data_path: Path) -> DataTuple:
...
+        assert len(s) == len(y), "data has to have the same length"
+        return cls(
+            data=pd.concat([s, y], axis="columns", sort=False),  # type: ignore[arg-type]
+            s_column=s_column,
+            y_column=y_column,
+            name=name,
+        )
+
+    @classmethod
+    def from_np(cls, *, s: npt.NDArray, y: npt.NDArray, s_name: str, y_name: str) -> LabelTuple:
```
MylesBartlett, line 321: Maybe make "s" and "y" the default names?

thomkeh: sure, good idea

ethicml/algorithms/inprocess/fair_dummies.py
```diff
@@ -52,7 +52,9 @@ def _get_path_to_script(self) -> List[str]:
 
     @implements(InAlgorithmSubprocess)
     def _get_flags(self) -> FairDummiesArgs:
-        model_type = "deep_model" if self.model_type.lower() == "deep_model" else "linear_model"
+        model_type: Literal["deep_model", "linear_model"] = (
+            "deep_model" if self.model_type.lower() == "deep_model" else "linear_model"
+        )
```
olliethomas, line 57: BTW, this is only here to satisfy the Literal type. For future work, the Literal should be replaced with an Enum.

olliethomas (approved):

ethicml/algorithms/inprocess/fair_dummies.py
thomkeh: should be easy to do

PR #670, opened 2022-05-28
Parallel pytest

thomkeh:


thomkeh:
Wow, this actually seems to make the tests twice as fast.

PR #674, opened 2022-05-30
make metrics part of evaluation run in parallel

olliethomas:
The obvious question, is why? Metrics are fast enough, what is there to be gained from parallelising this. The reason that I've done this is because there are some metrics will take longer (e.g. [this](https://github.com/yromano/fair_dummies/blob/master/fair_dummies/utility_functions.py#L347) trains a NN as part of the metric).

### DIFF
--- a/ethicml/evaluators/evaluate_models.py
+++ b/ethicml/evaluators/evaluate_models.py
@@ -1,9 +1,11 @@
 """Runs given metrics on given algorithms for given datasets."""
+import warnings
 from pathlib import Path
 from typing import Dict, List, NamedTuple, Optional, Sequence, Union
 from typing_extensions import Literal
 
 import pandas as pd
+from joblib import Parallel, delayed
 
 from ethicml.data.dataset import Dataset
 from ethicml.data.load import load_data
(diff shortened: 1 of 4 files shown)


ethicml/evaluators/evaluate_models.py
```diff
@@ -311,38 +328,61 @@ def _gather_metrics(
...
+        _run_metrics(model, data_info, predictions, metrics, per_sens_metrics)
+        for preds_for_dataset, data_info in zip(all_predictions_t, test_data)
+        for predictions, model in zip(preds_for_dataset, inprocess_models)
+    )
+
+    results_df = pd.DataFrame(columns=columns).from_dict(results)
+    if len(results_df):
+        all_results.append_df(results_df)
 
+    for _, data_info in zip(all_predictions_t, test_data):
```
thomkeh, line 342:
```suggestion
    for data_info in test_data:
```

ethicml/evaluators/evaluate_models.py
```diff
@@ -311,38 +328,61 @@ def _gather_metrics(
...
         # write results to CSV files and load previous results from the files if they already exist
         csv_file = _result_path(outdir, data_info.dataset_name, data_info.transform_name, topic)
-        aggregator = ResultsAggregator(results_df)
+        aggregator = ResultsAggregator(
+            results_df[
```
thomkeh, line 346:
```suggestion
            results_df.loc[
```

tests/visualisation_test.py
```diff
@@ -89,11 +89,11 @@ def test_plot_evals():
 
     # plot with metrics
     figs_and_plots = plot_results(results, Accuracy(), ProbPos())
-    # num(datasets) * num(preprocess) * num(accuracy combinations) * num(prop_pos combinations)
-    assert len(figs_and_plots) == 2 * 2 * 1 * 2
+    # num(datasets) * num(preprocess) * num(accuracy in results) * num(prob_pos in results)
+    assert len(figs_and_plots) == 1 * 2 * 1 * 2
```
thomkeh, line 93: This worries me that reading previous results is broken.

PR #688, opened 2022-06-14
Change dependencies for building docs

thomkeh:
Previously, ipython was a dependency, but that was overkill.


thomkeh:
Didn't help as much as I had hoped.

PR #692, opened 2022-06-22
When constructing a datatuple check for overlapping columns

thomkeh:


thomkeh:
Okay, I reworked the solution. Overlap between s and x is allowed now but only if the columns are actually identical. overlap between y and x is still not allowed (because that feels like a bug).

PR #693, opened 2022-06-22
Change the return type of `run_metrics()`

### LINKED ISSUE #690: run_metrics() should not append predictions.info to the results
This should be done somewhere else because not everyone using run_metric wants that behavior.

tmke8:
By moving the code that appends prediction.info out of the function.

closes #690


MylesBartlett (approved):

thomkeh:
apparently I was the one who added this in #263

thomkeh:
I'm kind of expecting this change will break something, but... eh.

PR #743, opened 2022-08-18
Fix two long-standing problems

### LINKED ISSUE #655: Fitted model on disk can be accidentally overwritten
Consider this case:

```python
>>> import ethicml as em
>>> adult_train, adult_test = em.train_test_split(em.data.Adult().load())
>>> from pathlib import Path
>>> ag1 = em.models.Agarwal(dir=Path("/tmp"))
>>> ag1.fit(adult_train)
>>> compas_train, compas_test = em.train_test_split(em.data.Compas().load())
>>> ag2 = em.models.Agarwal(dir=Path("/tmp"))
>>> ag2.fit(compas_train)
>>> ag1.predict(adult_test)
Failure: ['-m', 'ethicml.implementations.agarwal', '{"mode":"predict","test":"/tmp/tmprd5neaze/test.npz","predictions":"/tmp/tmprd5neaze/predictions.npz","model":"/tmp/model_Agarwal, lr, dp, 0.1.joblib"}', '{"classifier":1,"fairness":1,"eps":0.1,"iters":50,"C":1.0,"kernel":""}']
/home/tmk/.cache/pypoetry/virtualenvs/ethicml-dzQunYke-py3.10/lib/python3.10/site-packages/sklearn/base.py:493: FutureWarning: The feature names should match those that were passed during fit. Starting version 1.2, an error will be raised.
Feature names unseen at fit time:
- age
- capital-gain
- capital-loss
- education-num
- education_10th
- ...
Feature names seen at fit time, yet now missing:
- age-cat_25 - 45
- age-cat_Greater than 45
- age-cat_Less than 25
- age-num
- c-charge-degree_F
- ...
```
The problem is that the second `fit()`, `ag2.fit(compas_train)`, overwrote the model file, `/tmp/model_Agarwal, lr, dp, 0.1.joblib`. You can of course prevent this by setting the directories to two different values, but you could forget that.

One potential solution is to add a hash of the dataset to the saved model filename.

### LINKED ISSUE #666: ModelType should be an Enum instead of a Literal type

thomkeh:
closes #655
closes #666


ethicml/implementations/adv_debiasing_method.py
```diff
@@ -33,7 +33,7 @@ def fit(train: DataTuple, args: AdvDebArgs, seed: int = 888) -> AdvDebiasingClas
         cost_pred=torch.nn.CrossEntropyLoss(),
         in_shape=len(train.x.columns),
         batch_size=args["batch_size"],
-        model_type=args["model_type"],
+        model_type=ModelType(args["model_type"]),
```
olliethomas, line 36: i always get thrown about when `Enum(value)` is needed and `Enum[value]` - though the tests would tell us pretty quickly if there were a problem 😅

olliethomas (approved):
LGTM

ethicml/implementations/adv_debiasing_method.py
thomkeh: the tests do indeed fail, so let me check

thomkeh: So, it's `Enum(value)` and `Enum[name]`, but I'm using `StrEnum` here, so the value is just the lower case version of the name.

So:
```python
class E(StrEnum):
    A = auto()
    B = auto()

# this works:
E("a")
# and this works:
E["A"]
```

PR #829, opened 2023-01-03
Update Zafar to Python 3

thomkeh:
Depends on this branch: https://github.com/wearepal/fair-classification/tree/python3


thomkeh:
I will have to switch from pipenv to pdm.

PR #834, opened 2023-01-16
Switch Kamishima to Python3

thomkeh:


thomkeh:
I also ran the slow tests on hydra and they all passed.

PR #932, opened 2023-08-11
Make compatible with mypy 1.5

tmke8:


tmke8:
Okay, I forgot to actually update mypy, but dependabot will do this, and now our code is compatible with mypy 1.5.

